PR #48, opened 2026-08-14
add dsh-pain-point-check to Development and Runtime

ICCuse:
Adds [dsh-pain-point-check](https://github.com/ICCuse/dsh-pain-point-check) to **Development & Runtime** in both READMEs.

An enforced pain-point-check guard for DeepSeek Harness: after two non-converged experiments on one problem it injects the three questions, denies non-investigative tool calls until they are answered in the reply text, and blocks same-direction retries. Tested end-to-end (8 tests) against a real agent loop with a scripted mock adapter. Tagged `dsh-plugin`. See also the official discussion: https://github.com/deepseek-ai/deepseek-harness/discussions/895


fkysly:
Merged — thanks! 🎉

If you'd like, add the badge to your README:
```markdown
[![awesome · DSH plugin](https://awesome-dsh-plugin.com/badge.svg)](https://awesome-dsh-plugin.com)
```

PR #51, opened 2026-08-14
Add dsh-portable-launcher to Development & Runtime

15828148:
Add [dsh-portable-launcher](https://github.com/15828148/dsh-portable-launcher) - a one-click portable Windows launcher for the dsh Web UI (auto-installs Node.js and dsh, CN mirror fallback, retries + resume). One line per the contributing guide, added to both README.md and the Chinese README.


fkysly:
Thanks — dsh-portable-launcher is genuinely useful for Windows users! However this list only collects plugins installable via `dsh plugin add` (a `package.json` with a `dsh.bundle` manifest, see [contributing](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/blob/main/contributing.md)), and a launcher script falls outside that scope, so we can't merge this one.

If you ever wrap DSH-related functionality as an installable plugin, we'd be glad to list it. 🙌

PR #180, opened 2026-08-14
Add dsh-bookmarks: per-reply bookmarks with notes/tags and a cross-session center

penguin-oo:
## What

Adds [penguin-oo/dsh-bookmarks](https://github.com/penguin-oo/dsh-bookmarks) to both READMEs under the Sessions & Messages category.

## The plugin

- Bookmark any finalized assistant reply (button sits beside the 👍/👎 feedback controls), with an optional note and tags.
- Cross-session bookmark center: full-text search, tag filter, inline editing, jump back to session, one-click Markdown export, Alt+B toggle.
- Host service validates targets against the persisted session log, derives a text snippet server-side, and uses per-item compare-and-set versions (multi-tab safe).
- Install: \dsh plugin --profile web add dsh-bookmarks\

Built and verified against dsh 0.1.0-rc.6 (web profile).


fkysly:
Merged — thanks! Cross-session bookmark center with Markdown export fills a real gap.

PR #198, opened 2026-08-14
Add dsh-ops-kit to Workflow & Automation

LeslieWylie:
Adds `LeslieWylie/dsh-ops-kit` to **Workflow & Automation**.

An evidence-first operating kit. Six read-only tools — capability catalog, staged workflow plans, packaged skill reader, bounded local memory search, repository release audit, release checklist — plus five packaged skills. It plans and audits; it never writes to a remote, never creates issues, never mutates a repository, and never reads credential-shaped paths.

### What I actually verified

Listing verification in this index is self-reported (CI here never installs or runs a plugin), so here is what I ran rather than what I believe:

- `npm test` — 8/8, including three regression tests written for the defects below.
- A **harness boot test** that loads the package into a real cordis `Context` alongside `dsh-tools`, `dsh-skill`, `dsh-fs-local`, and two independent first-party control plugins, then executes every tool against real state — a real `git init`'d scratch repo, real files on disk, and a byte-comparison of `skill_read`'s output against the packaged `SKILL.md`. It is confirmed executing in CI, not skipping.
- Installed from a clean directory and loaded into a profile.

CI green at `c4ba835`.

### Three defects found and fixed while preparing this

1. **A path that *was* the configured root got rejected.** The default root came from `process.cwd()`, which Node reports with symlinks already resolved, while the caller's argument was only `resolve()`d — so on macOS `/tmp/x` vs `/private/tmp/x` failed containment against itself.
2. **One common keyword returned 20,261 bytes.** 20 files x 8 snippets x 300 chars, pretty-printed, crowding out the very context the answer was meant to inform. Now 5,133 bytes under an explicit byte budget, with `truncated` set honestly.
3. **A symlink inside a root that pointed outside it passed the containment check.** A comment claimed not following symlinks was deliberate. I measured it instead — root `/safe`, symlink `/safe/escape -> /outside`: `resolve()` alone judged `/safe/escape` to be *inside* `/safe`, letting a read escape the declared boundary. Both sides are canonicalized now, and the measurement is recorded next to the assertion.

### One thing worth flagging up front

The package name is `@dsh-community/dsh-ops-kit`. **That scope is unclaimed on npm — we do not own it, and nothing has ever been published under it.** The plugin installs from GitHub (`github:LeslieWylie/dsh-ops-kit`), which is what the README documents. I'm raising this rather than leaving you to find it, since the scope question came up on #136. Happy to rename if you'd prefer the entry not carry a scope we don't hold; two of my own local profiles key off it, so I'd change those in step.

Not a duplicate of my listed `dsh-fleet-audit`, which is a single credential-hygiene tool.


fkysly:
Merged — thanks for the rework! Self-contained now (only shells out to git), read-only by design, and the description matches what the code does. Exactly what we asked for — appreciated.

fkysly:
Hi @LeslieWylie 👋 你收录的插件现在可以在插件市场（[dsh-market](https://github.com/dsh-market/dsh-market) 1.8.0+）里像 App Store 一样展示截图了。

如果愿意，欢迎提个小 PR 把 1-8 张截图加进 [`data/screenshots.json`](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/blob/main/data/screenshots.json)——key 用你条目的 GitHub URL，图片放你自己仓库里即可（[说明](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/blob/main/contributing.md#screenshots--截图optional-recommended--可选推荐)）。**完全可选**：不提交的话市场会自动从你的 README 抽图，提交只是让你控制展示的顺序和内容。

/ Your plugin can now show AppStore-style screenshots in dsh-market (1.8.0+). If you like, open a small PR adding 1-8 image URLs to [`data/screenshots.json`](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/blob/main/data/screenshots.json) (keyed by your entry URL — [how](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/blob/main/contributing.md#screenshots--截图optional-recommended--可选推荐)). Entirely optional: markets fall back to your README images; a curated entry just gives you control over selection and order.

PR #299, opened 2026-08-15
test: canary for pr-check workflow (do not merge)

fkysly:
Functional test of `.github/workflows/pr-check.yml`. Commit 1 adds an entry to README.md only — the build must fail on locale parity. Commit 2 adds the zh line — the check must pass. Will be closed without merging.


fkysly:
Both paths verified: EN-only commit failed on locale parity, the zh fix passed (28s / 15s). The check works; closing unmerged.

PR #339, opened 2026-08-15
docs: add Catppuccin-dsh-theme and Solarized-dsh-theme

zhijun-dai:
Two theme plugins for the DSH Web theme runtime, both declaring dsh.bundle and published to npm.


zhijun-dai:
Thank you!!!

PR #386, opened 2026-08-15
Add dsh-movein-permissions, now dsh plugin add installable (follow-up to #363)

sjh9714:
Both fixes from the #363 review shipped in 0.2.3. plugin/cordis.patch.yml added, plugin/package.json declares dsh.bundle with the patch in files, verified end to end on a clean profile, dsh plugin --profile web add dsh-movein-permissions registers the bundle and dsh web boots with zero module errors (the bundle row is config-less and inert until the user adds deny/ask rules, so a bare install cannot break a boot).

On the one-entry-or-two question, keeping it separate is deliberate, the gate works standalone for DSH users who never migrate (write the rules by hand), and movein users get the row generated for them either way. Separate entries match the two audiences.


fkysly:
已合并 🎉 补齐得很快。你这两天走完两轮"退回 → 修好 → 上架"（dsh-movein 本体 + 这个权限子包），效率很高。

PR #746, opened 2026-08-15
Add dsh-result-only-view to UI Enhancements

YEYEYEYESHIFU:
Adds a Results-only conversation toggle for the DeepSeek Harness Web GUI.

- npm: \dsh-result-only-view\ (v1.4.0, MIT)
- Repo: https://github.com/YEYEYEYESHIFU/dsh-result-only-view
- Hides thinking rows and tool-call nodes while the agent works, keeps one live status line for activity feedback, restores working animations under prefers-reduced-motion, zh-CN/en localized; ask_user_question and cordis_run cards stay visible.

Added one entry to \README.md\ and \README.zh.md\ under the UI Enhancements / 🎨 UI 增强 category per the contributing guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,6 @@ dsh plugin --profile web add dshmarket
 ## Plugins
 
 ### UI Enhancements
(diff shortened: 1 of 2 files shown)


YEYEYEYESHIFU:
Moved the entry next to the existing auto-collapse plugins (renat3u/dsh-web-archive, a179-sanae/dsh-auto-collapse) that share the same use case, instead of pinning it to the top of the UI Enhancements category.

PR #821, opened 2026-08-16
Move `dsh-context` to Memory

bowenliang123:
Move the `dsh-context` entry (context insight panel) from the "UI Enhancements" section to the **first position** of the "Memory" section, in both `README.md` and `README.zh.md`.

**Rationale:** the repo's "Memory" section hosts context-lifecycle plugins (context pruning / compression, e.g. `aerince/dsh-active-context-pruning` and `Tyan66666/billion-context-dsh`). `dsh-context` inspects the model's context window — composition vs. window size, per-request history, compression/injection events, and per-message token stats — so it belongs in Memory.

- **License:** Apache-2.0
- **Description:** Context insight panel: see what the model's context window is made of and how it evolves — composition vs. window size, per-request history, compression/injection events, and per-message token stats.
- **Install:** `dsh plugin --profile web add dsh-context`

Only the `dsh-context` line was moved; no other entries were touched.


bowenliang123:
Thanks for merging this. But I'm thinking about moving it back to UI category, since my plugin is more about UI enhancement after all. I'll raise another PR for that. Thanks for the patience!

PR #947, opened 2026-08-16
Add nzl153/dsh-preflight to Development & Runtime

nzl153:
Adds one entry under `Development & Runtime` in both `README.md` and `README.zh.md`.

`dsh-preflight` is a CLI with two halves:

- **Before installing** — fetch a candidate plugin (npm / GitHub ref / local path), unpack it, and collide it against the current profile: entry-id collisions, duplicate service providers, silent config overrides, peer/version drift, and entries missing from the published artifact.
- **After something breaks** — read the DSH logs and reduce a screenful of cascading plugin errors to a single root cause (port already in use, bundle unresolved at boot, entry id registered twice, missing module entry).

**Read-only by design.** It downloads and statically parses JSON/YAML and source text; it never installs a plugin, never executes candidate code or install scripts, and never modifies a profile. Suggested fixes are printed as commands for the user to run. Network candidates are unpacked into a temp dir with traversal/symlink/size/file-count limits and removed afterwards.

The README documents which rules have been verified against reproduced real failures and which are unit-tested only, plus the known limits — the log patterns are written against the current DSH log wording, and unmatched errors degrade to an explicit UNKNOWN finding with the raw lines attached rather than reporting all-clear.

MIT, Node >= 22.19, Windows-first.


nzl153:
Closing this one myself — thanks for the clear CI message.

`dsh-preflight` is a standalone CLI rather than an installable plugin: it has no `dsh.bundle` manifest and no `cordis.patch.yml`, because it runs outside DSH and only reads profiles and logs. So it doesn't meet the submission bar, and adding a bundle manifest just to qualify would misrepresent what it is.

It will live in its own repo instead. Sorry for the noise.

PR #1316, opened 2026-08-17
add a903067276-rgb/dsh-perm-guard (auto-approval permission guard)

a903067276-rgb:
Add [dsh-perm-guard](https://github.com/a903067276-rgb/dsh-perm-guard) — an auto-approval permission guard plugin for DSH: a middle tier between `workspace-write` (asks too often) and `danger-full-access` (too open). Auto-allows safe operations inside trust directories, always asks a human for destructive ones; 11 per-category tri-state switches, two modes (Standard / Aggressive), command-level firewall and an audit trail. Category: workflow.

- Repo declares `dsh.bundle` manifest ✅
- Bilingual README, MIT license, install docs, UI screenshots ✅
- Tagged v0.1.1 (security fix: trust-boundary traversal — path normalization via resolve + realpathSync, unit-tested 7/7 and live-verified) ✅
- Verification matrix: `docs/verify-checklist.md` (standard + aggressive full-category tests) ✅


fkysly:
Thanks — this is a well-built plugin and the reasoning behind it is sound. The rule set, the command classification, the fail-safe default for unknown commands, the audit trail, the loopback-only state endpoint, and your own `docs/verify-checklist.md` all read like someone who took the problem seriously. Two things block it, though, and the second one matters a lot given what this plugin is for.

**1. The trust boundary can be walked out of.**

`inTrust()` in `lib/index.js` compares strings:

```js
const normPath = (p) => { let s = p.trim(); if (s.length > 1 && s.endsWith('/')) s = s.slice(0, -1); return s }
const inTrust = (target, roots) => {
  const t = normPath(target)
  if (!t.startsWith('/')) return false
  for (const r of roots) { const root = normPath(r); if (t === root || t.startsWith(root + '/')) return true }
  return false
}
```

`normPath` trims whitespace and a trailing slash. It does not resolve `..`, and nothing calls `path.resolve` or `fs.realpathSync`. So:

```
/Users/me/proj/../../../etc/cron.d/x
```

literally starts with the trust root `/Users/me/proj/`, so `inTrust()` returns `true` — while the OS resolves it far outside the workspace.

The `PROTECTED` list doesn't catch it either:

```js
/(^|\s)\/(etc|usr|System|Library|Applications)(\/|\s|$)/
```

that requires whitespace or start-of-string before `/etc`; here the preceding character is `.` from `..`, so it doesn't match. Telling detail: the `.ssh`/`.aws` pattern only requires a literal `/` before it, so traversal *is* caught there — the two patterns disagree with each other.

This reaches both auto-approve paths — the `write`/`edit` branch and the bash target check. A path with `../` gets silently rubber-stamped instead of asking. For a plugin whose entire promise is "auto-allows safe operations inside trust directories, always asks a human for destructive ones", that inverts the guarantee — and the caller is an LLM, which may be acting on a prompt-injected instruction.

The fix is small: resolve both sides and compare the resolved values.

```js
import path from 'node:path'
import fs from 'node:fs'

const realOf = (p) => { try { return fs.realpathSync(path.resolve(p)) } catch { return path.resolve(p) } }
const inTrust = (target, roots) => {
  const t = realOf(target)
  return roots.some((r) => { const root = realOf(r); return t === root || t.startsWith(root + path.sep) })
}
```

`realpathSync` matters as much as `resolve` — a symlink inside the workspace pointing outward defeats `resolve` alone. Falling back to `resolve` when the path doesn't exist yet keeps new-file writes working.

**2. The repo is under the age bar.** It was created about 3.5 hours before this PR; entries need **1 day and 10+ commits**. You already have the commits. Resubmit once it ages in — nothing is held against a resubmission.

Worth saying: our gate reported this PR as passing, which was wrong — a bug on our side that I've since fixed. The age check never actually ran on your entry, and the check-run claimed it had. Sorry for the mixed signal.

Fix the path normalisation and come back after the repo clears a day, and I'll re-review. Happy to look at the patch before you resubmit if that's useful.

---

感谢提交——这个插件做得不错,思路也是对的。规则集、命令分类、未知命令的 fail-safe 默认、审计记录、只监听本地回环的状态接口,以及你自己的 `docs/verify-checklist.md`,都能看出是认真对待这个问题的。但有两点阻塞,其中第二点对这个插件尤其关键。

**1. 信任边界可以被走出去。**

`lib/index.js` 里的 `inTrust()` 是纯字符串比较(见上)。`normPath` 只去空格和尾部斜杠,**不折叠 `..`**,全文也没有 `path.resolve` 或 `fs.realpathSync`。于是 `/Users/me/proj/../../../etc/cron.d/x` 在字面上以信任根 `/Users/me/proj/` 开头,`inTrust()` 返回 `true`,而操作系统解析出来的位置早已在工作区之外。

`PROTECTED` 也拦不住:那条正则要求 `/etc` 前面是空白或字符串开头,而这里前面是 `..` 的 `.`,匹配不上。一个能说明问题的细节是——`.ssh`/`.aws` 那条只要求前面有 `/`,所以**穿越在那里反而能被拦下**,两条规则自相矛盾。

这条路径同时影响 `write`/`edit` 分支和 bash 目标检查:带 `../` 的目标会被静默放行,而不是询问人。对一个承诺"信任目录内自动放行、破坏性操作必问人"的插件,这把保证反过来了——而且调用方是 LLM,它可能正在执行一条被注入的指令。

修法很小:两边都解析后再比较(代码见上)。`realpathSync` 和 `resolve` 一样重要——工作区内一个指向外部的符号链接,单靠 `resolve` 是拦不住的;路径尚不存在时回退到 `resolve`,可以保证新建文件的写入不受影响。

**2. 仓库未达年龄门槛。** 它在此 PR 前约 3.5 小时创建,而收录要求**满 1 天且提交数 ≥ 10**。提交数你已经够了,等仓库满一天后重新提交即可,重提不会有任何影响。

另外要说明:我们的 gate 把这个 PR 判成了通过,**那是我们这边的 bug,现已修复**——年龄检查实际上从未在你的条目上运行,而 check 却声称它运行了。给你造成的信号混乱,抱歉。

把路径规范化修掉、仓库满一天后回来,我会重新审。如果需要,提交前我也可以先帮你看一眼补丁。

a903067276-rgb:
Thanks for the review — the traversal finding was correct and serious, and it is now fixed and shipped as **v0.1.1**.

## Fix (per your suggested approach)

`inTrust()` now normalizes both sides before comparing:

```js
const expandHome = (p) => p === "~" ? homedir() : p.startsWith("~/") ? join(homedir(), p.slice(2)) : p
const realOf = (p) => {
  const abs = resolve(expandHome(p.trim()))
  try { return realpathSync(abs) } catch { return abs } // non-existent target (new file) falls back to resolve
}
const inTrust = (target, roots) => {
  const t = realOf(target)
  if (!t.startsWith("/")) return false
  for (const r of roots) {
    const root = realOf(r)
    if (t === root || t.startsWith(root + sep)) return true
  }
  return false
}
```

- `resolve` collapses `..`; `realpathSync` defeats symlinks pointing outside the workspace; non-existent paths (new-file writes) fall back to `resolve`.
- Both auto-approve paths are covered: the bash target check now also re-runs the PROTECTED patterns against the **resolved** path (standard and aggressive modes), and the `write`/`edit` branch checks both the raw and the resolved path.
- Also hardened: `~`-prefixed targets are expanded before comparison.

Verified with a 7-case unit test (traversal attack, `../../../../etc/passwd`, normal workspace file, subdirectory, new-file fallback, sibling dir, `~` expansion) plus a live test through the running plugin: the exact `…/trust/../../../Documents/…` shape now yields `ask-human/outside` and prompts the human instead of being silently approved.

## Age gate

Understood — the repo is under the 1-day age bar. I will resubmit once it clears (currently 14 commits, well above the 10-commit requirement). Happy to re-run the gate at that point.

Thanks again for catching it.

a903067276-rgb:
The repo has now aged past the requirement and the gate re-run is green on the latest commit:

- **Submission gate: success**
- **check: success**

PR description updated to reflect **v0.1.1** (the traversal security fix: `resolve` + `realpathSync` path normalization, unit-tested 7/7 and live-verified — write targeting `…/trust/../../../Documents/…` now prompts instead of being auto-approved). Ready for re-review, thanks!

fkysly:
Holding this one, and I want to be straight about why it gets a harder look than most: an auto-approval plugin is the one category where the whole value is "you can stop reading the prompts". That makes any way for it to widen its own permissions a different class of problem from the same bug in an ordinary plugin.

The build is genuinely careful, and I want that on the record before the findings. `loadConfig()` reverts to Standard-mode defaults on a missing or malformed config rather than to something more open. With no sandbox or session context it defers to the host via `next()` instead of deciding. The v0.1.1 traversal fix is real: `inTrust()` compares two `realpathSync`-resolved absolute paths with a separator-anchored prefix check, which closes both `../..` traversal and symlink evasion for paths that exist. The eleven category switches are eleven keys in `CATEGORY_DEFAULTS`, exactly as the entry says. Nothing leaves the machine.

Two findings. I read `lib/index.js` myself rather than relying on a summary.

## 1. In Aggressive mode the plugin can be talked into widening itself

`classifyCommand()` ends at line 242:

```js
if (mode === 'aggressive') return { d: 'allow', c: 'unknown' } // 激进档：未知命令全放行
```

Anything the classifier does not recognise is auto-allowed with no prompt. `curl` is not in `WRITE_CMDS`, is not git or npm, and does not match DANGER or PROTECTED — so it reaches line 242 and is allowed. Note that the `curl`/`wget` → `networkExec` rule three lines below at 245 is unreachable in this mode, because 242 returns first.

The config endpoint accepts a patch:

```js
if (body.categories && typeof body.categories === 'object') { ... saveConfig(patch) }
```

guarded only by `isSameOrigin(req)`, which tests the **Host header** against `127.0.0.1|localhost`. A command running on the same machine satisfies that by construction.

So in Aggressive mode a single unprompted command can set `delete` and `protected` to `auto`. A later `rm -rf ~/some/real/project` — not the bare `/` or `~` the DANGER regex catches — then classifies as `delete → auto` and runs with no confirmation. That contradicts the README's "Never auto-approved (all modes): … Writes to protected paths".

**Standard mode is not affected** — its unknown-command fallback is `ask`. This only fires once a user has opted into Aggressive, which matters for severity but not for the principle: a permission gate that a shell command can reconfigure is not a gate in the mode where it counts.

## 2. The traversal fix does not cover paths that do not exist yet

Line 141:

```js
try { return realpathSync(abs) } catch { return abs } // 路径尚不存在（新建文件）时回退 resolve
```

The fallback is correct in intent — you cannot realpath a file you are about to create — but it means the trust check for **new** paths is a literal string comparison. Two steps:

1. `ln -s /etc /workspace/evil` — the link's own destination does not exist yet, so it is judged in-trust by string prefix and allowed.
2. Write to `/workspace/evil/x` — ENOENT again, string prefix again, allowed. The file is created under `/etc`.

Both steps stay inside the trusted root as far as the check can see. This is the case the v0.1.1 fix was aimed at, arriving by a different door.

## What would unblock it

1. In Aggressive mode, route unrecognised or shell-complex commands to `ask` rather than `allow`. "Aggressive" can reasonably mean a wider auto-allow list; it should not mean auto-allowing things the classifier could not parse, since that set is unbounded by definition.
2. In `realOf()`, walk up to the deepest **existing** ancestor, `realpathSync` that, and rejoin the remaining segments — so a new file under a symlinked parent is judged by where it will actually land.

Both are contained changes and neither touches the design. I would like to list this — the category taxonomy, the tri-state switches and the fail-safe config load are better thought through than most in this space, which is exactly why the two holes are worth closing rather than shipping around.

Category `workflow` is fine; I am not asking you to change it.

---

暂缓合并。先直说为什么它会被更严格地审视:自动批准类插件是唯一一种「全部价值就在于你可以不再逐条看提示」的插件,因此**任何能让它自己放宽权限的路径**,其性质都与普通插件里的同类 bug 不同。

这个实现确实是用心的,findings 之前我想先把这点记录下来:`loadConfig()` 在配置缺失或损坏时回退到 Standard 档默认值,而不是更开放的状态;没有 sandbox/session 上下文时通过 `next()` 交还宿主而不是自行决定;v0.1.1 的穿越修复是真的——`inTrust()` 比较的是两个经 `realpathSync` 解析的绝对路径,并做了以分隔符为界的前缀比较,对**已存在**的路径,`../..` 穿越和符号链接规避都被堵住了;十一个类别开关确实是 `CATEGORY_DEFAULTS` 里的十一个键,与条目描述一致;没有任何数据离开本机。

两个发现。我自己读的 `lib/index.js`,不是采信摘要。

## 1. 激进档下,插件可以被诱导放宽自身权限

`classifyCommand()` 结尾第 242 行:`if (mode === 'aggressive') return { d: 'allow', c: 'unknown' } // 激进档：未知命令全放行`。

分类器不认识的一切都会被无提示自动放行。`curl` 不在 `WRITE_CMDS` 里,不是 git/npm,也不匹配 DANGER/PROTECTED——因此它会走到第 242 行被放行。注意:三行之后第 245 行那条 `curl`/`wget` → `networkExec` 规则在这个模式下**永远到不了**,因为 242 先返回了。

而配置端点接受 patch 写入,唯一的守卫是 `isSameOrigin(req)`——它检查的是 **Host 头**是否匹配 `127.0.0.1|localhost`。一条在同一台机器上运行的命令天然满足这个条件。

于是在激进档下,**一条无需提示的命令就能把 `delete` 和 `protected` 改成 `auto`**。随后一条 `rm -rf ~/某个真实项目`(而不是 DANGER 正则拦的裸 `/` 或 `~`)就会被归类为 `delete → auto` 并直接执行,不再确认。这与 README 里「所有模式下永不自动批准:……写入受保护路径」相矛盾。

**标准档不受影响**——它的未知命令回退是 `ask`。这只在用户主动切到激进档后才触发,这影响严重性,但不改变原则:**一个能被 shell 命令重新配置的权限闸门,在最需要它的那个模式下就不是闸门。**

## 2. 穿越修复没有覆盖「尚不存在」的路径

第 141 行:`try { return realpathSync(abs) } catch { return abs } // 路径尚不存在（新建文件）时回退 resolve`。

这个回退的意图是对的——你无法 realpath 一个即将创建的文件——但它意味着对**新建**路径的信任判断退化成了字面字符串比较。两步即可:

1. `ln -s /etc /workspace/evil` —— 链接自身的目标尚不存在,按字符串前缀判为可信,放行。
2. 写入 `/workspace/evil/x` —— 又是 ENOENT,又是字符串前缀,放行。文件实际创建在 `/etc` 下。

两步在检查看来都待在受信根目录内。这正是 v0.1.1 想修的那类问题,只是换了一扇门进来。

## 解除条件

1. 激进档下,把无法识别或 shell 结构复杂的命令路由到 `ask` 而不是 `allow`。「激进」可以合理地意味着更宽的自动放行清单;但不应意味着自动放行那些**分类器根本没能解析的东西**——那个集合按定义是无界的。
2. `realOf()` 里向上找到最深的**已存在**祖先目录,对它做 `realpathSync`,再把剩余路径段拼回去——这样一个位于符号链接父目录下的新文件,会按它**实际落点**来判断。

两处都是局部改动,都不触及设计。这个插件我是希望收录的——类别划分、三态开关、以及配置加载的 fail-safe,都比这个领域里的多数实现考虑得更周到,这恰恰是这两个洞值得补上、而不是绕过去的原因。

分类 `workflow` 没问题,我不要求你改。

a903067276-rgb:
Thank you for the review — both findings are real and both are now fixed, plus one extra hardening step that closes the 'widen its own permissions' path completely.

## Finding 1 — fixed in v0.1.4 + v0.1.5
- **v0.1.4**: Aggressive mode no longer auto-allows commands the classifier cannot parse — unknown/complex commands route to `ask`. Explicit categories (networkExec/gitPush/publish) still auto-allow per user config.
- **v0.1.5 (extra)**: high-risk categories — `danger` / `protected` / `delete` / `disk` / `privilege` — are now **locked to `ask` in every mode**. Config patches that try to set them to `auto` or `deny` are rejected at load, at save, and at the HTTP endpoint. The `rm -rf ~/real/project` after reconfiguring `delete` to auto is no longer reachable by any route.

## Finding 2 — fixed in v0.1.4
`realOf()` now walks up to the deepest existing ancestor, realpaths it, and rejoins the remaining segments — a new file under a symlinked parent (`ln -s /etc <trusted>/evil` → write `<trusted>/evil/x`) is judged by its actual landing point (/private/etc/x on macOS). Verified with a live symlink test.

Also in v0.1.5: wider command recognition (jq/sha256sum/ping/dig read-only, interpreters judged by trust-dir script vs arbitrary -c/-e code, scp/rsync → network, kill/xargs → locked privilege, xargs+rm danger flag) — fewer false unknowns means auto-approval only ever covers things the classifier genuinely understands.

Releases: https://github.com/a903067276-rgb/dsh-perm-guard/releases/tag/v0.1.5

### 中文说明
感谢审阅，两个 finding 均已修复并加固：激进档未知命令改人工确认（0.1.4）；高危类别（danger/protected/delete/disk/privilege）任何模式强制 ask、配置防篡改（0.1.5，彻底封死自放宽路径）；symlink 父目录新建文件按实际落点判定（0.1.4）。另扩充常用命令识别，减少误判。请复查。

a903067276-rgb:
A personal note, if I may: I'm not a developer by trade — I build with AI agents and learn as I go. Having someone actually read the code and take it apart the way you did (why an auto-approval plugin deserves a harder look than most, why a permission gate a shell command can reconfigure is not a gate, judging a new file by where it will really land) has been genuinely educational, and I mean that. Thank you for reading the source yourself instead of trusting a summary — it made this a much better plugin.

### 中文说明
说点题外话：我是非从业者，平时靠 AI 帮我写东西、边做边学。你能逐行读代码、把问题拆得这么清楚（自动批准类插件为什么该被更严格审视、'能被 shell 命令重配的权限闸门不是闸门'、'新建文件要按实际落点判定'），每一条都让我受益匪浅。谢谢你亲自读源码而不是采信摘要——因为这个插件变好了很多。

fkysly:
Reclassified from `workflow` to `security` and merging.

Permission gating with an audit trail is what the `security` section is for — `workflow` is task automation. One-line change made on your branch; nothing needed from you.

I read `lib/index.js` closely because auto-approval is the kind of thing worth being sure about, and the design holds up:

- `LOCKED_ASK` (`delete`, `protected`, `disk`, `privilege`, `danger`) is enforced in three independent places — `catOf()`, `saveConfig()`, and the HTTP state handler — so a hand-edited or corrupted config file can't elevate a destructive category to auto.
- An unresolvable workspace root, or a tool call that can't be located in session events, falls through to `next()`. It fails to *ask*, not to *allow*.
- `realOf()` resolving symlinks and `../` before the trust check is the right call, and the local-CSRF guard on the config endpoint is more than most plugins bother with.
- "11 per-category switches" is exact — I counted `CATEGORY_DEFAULTS`.

**One thing to fix in your repo** (not blocking, not part of the entry): your README's Coverage section still says unknown commands fall back to `allow` in Aggressive mode. The code no longer does that — every `unknown` branch returns `ask` regardless of mode, and there's a comment marking the removal. The code is safer than the doc claims, which is the good direction for them to disagree in, but someone reading the README to decide whether to trust this will get the wrong answer.

---

已从 `workflow` 改为 `security`，随后合并。

带审计记录的权限管控正是 `security` 这一节的用途，`workflow` 指的是任务自动化。这一行改动已在你的分支上做好，你无需操作。

自动放行这种事值得看仔细，所以我把 `lib/index.js` 通读了一遍，设计是站得住的：

- `LOCKED_ASK`（`delete`、`protected`、`disk`、`privilege`、`danger`）在三个互相独立的位置强制执行——`catOf()`、`saveConfig()` 和 HTTP state 处理器——所以手改或损坏的配置文件无法把危险分类提升为自动放行。
- 工作区根目录解析不出来、或在 session 事件里找不到对应的工具调用时，会落到 `next()`。失败时是**询问**，不是**放行**。
- `realOf()` 在信任判断前先解析软链接和 `../`，这个处理是对的；配置端点上的本地 CSRF 防护也比大多数插件做得多。
- 「11 个分类开关」数字准确——我数过 `CATEGORY_DEFAULTS`。

**你仓库里有一处建议修**（不阻塞合并，也不属于条目内容）：README 的 Coverage 一节仍写着 Aggressive 模式下未知命令回退为 `allow`。代码已经不是这样了——所有 `unknown` 分支无论什么模式都返回 `ask`，而且有注释标明这个逻辑被删掉了。代码比文档更安全，两者不一致时这是好的方向；但有人靠 README 判断要不要信任这个插件时，会得到错误的答案。

a903067276-rgb:
Thank you for the reclassification and the merge — and especially for reading the code that carefully. Being told "the design holds up" by someone who actually checked is worth more than any badge.

**README fixed** (commit 4139ff2): the Coverage section now states that unknown commands always fall back to ask regardless of mode — matching the code. Good catch; docs should never claim the code is looser than it is.

Looking forward to the security section neighbors!

### 中文说明
感谢重新分类与合并，更感谢你逐行读了代码——「设计站得住」这句话比任何徽章都值钱。README 的 Coverage 一节已同步代码现状（未知命令无论模式一律 ask，提交 4139ff2），文档不会再比代码更宽松了。很高兴入驻 security 分区！

PR #1066, opened 2026-08-16
Add deepseek-ai/deepseek-harness to the curated list

xigua6661999:
Adds the official repository [deepseek-ai/deepseek-harness](https://github.com/deepseek-ai/deepseek-harness) to the Development & Runtime category.

Eligibility:

- Declares a `dsh.bundle` manifest in subpackages (packages/bundle/base, packages/bundle/headless, packages/bundle/web-app), qualifying as a monorepo.
- Already carries the `dsh-plugin` GitHub topic.
- Active, long-lived project (far above the 1-day / 10-commit bar).

Both READMEs regenerated via `node scripts/generate-readme.mjs` (985 entries).

PR #1339, opened 2026-08-17
Refuse DeepSeek Harness itself as a submission

Chinesezjc:
## What

`check-submission.mjs` refuses `deepseek-ai/deepseek-harness` by identity, before the manifest and threshold checks run.

## Why

The harness satisfies every current rule, so the gate would accept the product as a plugin for the product:

- `packages/bundle/base/package.json` is `@deepseek-ai/dsh-base` and declares `dsh.bundle`
- it is the **19th** of **248** `package.json` files in the tree, so the enumeration reaches it well inside `MAX_TREE_PKGS = 40`
- `truncated` is `false`, so the tree is fully readable and no "unknown" path is taken
- age and commit thresholds pass by years

The contract checks are correct here — the harness genuinely declares a bundle. What makes it wrong as an entry is identity, not conformance, so that is where the check goes.

It is not currently listed; this closes the path before someone submits it.

## How to verify

Reproduce the accept-path facts:

```sh
gh api "repos/deepseek-ai/deepseek-harness/git/trees/HEAD?recursive=1" \
  --jq '{truncated, manifests: [.tree[]|select(.path|endswith("package.json"))]|length}'
# => {"truncated": false, "manifests": 248}

gh api repos/deepseek-ai/deepseek-harness/contents/packages/bundle/base/package.json \
  --jq '.content' | base64 -d | jq -c '{name, dsh}'
# => {"name":"@deepseek-ai/dsh-base","dsh":{"bundle":{"patch":"./cordis.patch.yml"}}}
```

Then confirm the guard rejects the harness while still accepting plugins, including a monorepo one and another repo in the same org:

| URL | Expected |
| --- | --- |
| `deepseek-ai/deepseek-harness` | refused |
| `deepseek-ai/deepseek-harness/tree/main/packages/bundle/base` | refused |
| `DeepSeek-AI/DeepSeek-Harness` | refused (case) |
| `Chinesezjc/dsh-interconnect` | accepted |
| `zhu1090093659/dsh-web-ui` | accepted (monorepo) |
| `deepseek-ai/some-other-repo` | accepted (same org) |

I checked this both ways: with the guard all six behave as above; with `FIRST_PARTY_REPOS` emptied, the three refusals turn into acceptances. Happy to add that as a committed test if you have a preferred location — `scripts/` has no test convention today, so I left it out rather than introduce one.

### DIFF
--- a/scripts/check-submission.mjs
+++ b/scripts/check-submission.mjs
@@ -5,6 +5,7 @@
 //      whole tree is enumerated rather than a guessed list of directories)
 //   2. the repo is at least MIN_AGE_DAYS old and has >= MIN_COMMITS commits
 //   3. the repo exists and isn't archived
+//   4. the repo is not DSH itself (it declares `dsh.bundle` and would pass 1-3)
 //
 // Needs GITHUB_TOKEN: the git-tree enumeration and the commit count are API
 // calls, and unauthenticated (60/hr per IP) is nowhere near enough. That is
@@ -21,6 +22,15 @@ const MIN_AGE_DAYS = 1
 const MIN_COMMITS = 10
 const CONCURRENCY = 6
 const MAX_TREE_PKGS = 40
+
+// DSH itself declares `dsh.bundle`: packages/bundle/base/package.json is
+// @deepseek-ai/dsh-base, and it is the 19th of 248 manifests in that tree, so
+// the enumeration reaches it well inside MAX_TREE_PKGS and the age and commit
+// thresholds are met by years. The harness would therefore pass the gate as a
+// plugin for itself. Listing the product in a list of plugins for the product
+// is the one wrong entry every visitor would recognise, so it is refused by
+// identity rather than by contract.
+const FIRST_PARTY_REPOS = new Set(['deepseek-ai/deepseek-harness'])
 
 // Entries submitted before the gate existed are judged by the old rules; only
 // the manifest check applies to them. Set to when the rule change landed.
@@ -131,6 +141,9 @@ async function commitCount(repo) {
 
 async function check(entry) {
   const { repo, sub } = decompose(entry.url)
+  if (FIRST_PARTY_REPOS.has(repo.toLowerCase())) {
+    return ['this is DeepSeek Harness itself, not a plugin for it']
+  }
   const meta = await api(`repos/${repo}`)
   if (meta.status === 404) return [`repository not found: https://github.com/${repo}`]
   if (meta.status !== 200) {


fkysly:
Merged — and thank you for wiring in the guard rather than just reporting it. This is the case from #1066, where someone submitted the harness itself and it was turned away by hand.

I verified the claims in your comment rather than taking them on trust, and they all hold: `packages/bundle/base/package.json` is `@deepseek-ai/dsh-base` and does declare `dsh.bundle`; it is the 19th of 248 manifests in that tree, so the enumeration reaches it well inside `MAX_TREE_PKGS = 40`. The harness really would have passed the gate as a plugin for itself. Short-circuiting before the API call is the right placement too.

One consequence worth naming, not a blocker: `decompose()` returns the same `repo` for a subpackage URL, so `deepseek-ai/deepseek-harness/tree/main/packages/<something>` is refused as well. Today that is almost certainly what we want — a bundle shipped inside the harness is part of the product, not a community plugin. But if first-party plugins are ever published from that monorepo, this will need to key on the full URL rather than the repo. Left as is; noting it so the next person reading the code knows it was a choice.

---

已合并——感谢你直接把守卫写好，而不是只报告问题。这正是 #1066 的情形：有人提交了 harness 本身，当时是人工拒掉的。

你注释里的事实断言我逐条核实过而非直接采信，结论全部成立：`packages/bundle/base/package.json` 确实是 `@deepseek-ai/dsh-base` 且声明了 `dsh.bundle`；它在该 tree 的 248 个 manifest 中排第 19 个，枚举远在 `MAX_TREE_PKGS = 40` 之内。harness 确实会作为「自己的插件」通过准入检查。在 API 调用前短路的位置也选得对。

有个副作用值得点出，不构成阻塞：`decompose()` 对子包 URL 返回相同的 `repo`，因此 `deepseek-ai/deepseek-harness/tree/main/packages/<某包>` 也会被一并拒绝。就现在而言这几乎肯定是我们想要的——随 harness 发布的 bundle 属于产品本身，不是社区插件。但如果将来有第一方插件从该 monorepo 发布，这里就需要改成按完整 URL 判断。暂不改动，写下来是为了让下一个读这段代码的人知道这是个选择，不是疏漏。

Chinesezjc:
Confirmed, and I should have named that in the description rather than leaving you to find it.

Verified the mechanism: `decompose()` takes the first two path segments, so both of these produce `repo: deepseek-ai/deepseek-harness` and the guard fires before `sub` is ever consulted.

```
https://github.com/deepseek-ai/deepseek-harness                              -> repo=deepseek-ai/deepseek-harness  sub=-
https://github.com/deepseek-ai/deepseek-harness/tree/main/packages/bundle/base -> repo=deepseek-ai/deepseek-harness  sub=packages/bundle/base
```

I agree with leaving it. Keying on the full URL would let `.../tree/main/packages/<x>` through, and every bundle currently in that tree is a product package — `dsh-base`, `dsh-web-app`, `dsh-headless` are what the harness itself mounts. Refusing them is the correct answer today, and a URL-keyed check would be a silent invitation to list them.

If first-party plugins are ever published from that monorepo, the signal to key on is probably narrower than the full URL: the product bundles are a known, small set, so an allowlist of subpaths would fail closed, whereas full-URL matching fails open. Noting that here rather than pre-building it.

Related: #1364 refuses repositories that *vendor* those same packages — six on the topic carry `@deepseek-ai/dsh-base` verbatim with `fork: false`, so they pass by containing the harness rather than by forking it. Different mechanism, same underlying confusion between the product and a plugin for it.

PR #21, opened 2026-08-14
Add dsh-sticky-disclosure (UI Enhancements)

Han-1413141:
Adds [dsh-sticky-disclosure](https://github.com/Han-1413141/dsh-sticky-disclosure) under UI Enhancements in both README.md and README.zh.md, and bumps the plugin count to 123.

It is a dsh-installable bundle (declares \dsh.bundle\ + \dsh.client\), carries the \dsh-plugin\ topic, and ships a Playwright test suite (48 assertions) plus bilingual docs.

PR #1364, opened 2026-08-17
Refuse repositories that vendor DSH's own bundle packages

Chinesezjc:
## What

`check-submission.mjs` refuses a repository whose `dsh.bundle` manifest is one of DSH's own bundle packages, unless a genuine bundle appears elsewhere in the tree.

## Why

Five repositories on the topic ship `packages/bundle/base/package.json` naming `@deepseek-ai/dsh-base` verbatim. They contain the harness rather than a plugin for it, and they pass the current gate:

| Repository | Stars | Created | `dsh.bundle` found at |
| --- | --- | --- | --- |
| `fufankeji/deepseek-harness-studio` | 208 | 2026-08-15 | manifest #21 of 255 |
| `KeepLost/harniverse` | 0 | 2026-08-14 | manifest #21 of 253 |

Both are inside `MAX_TREE_PKGS = 40`, both have `truncated: false`, and both clear the age and commit thresholds. `soolaugust/deepseek-harness-cli`, `my-dsh/oh-my-dsh` and `BenHuHuan/dhs-tuicode` carry the same package.

`fork` is **false** for all five — these are source copies, not GitHub forks, so an owner check or a fork check does not see them. The name is not the author's mistake; it is the harness's own name, which is why this is decided by package identity rather than by contract.

None is currently listed. This closes the path rather than correcting an entry.

## How to verify

```sh
gh api "repos/fufankeji/deepseek-harness-studio/git/trees/HEAD?recursive=1" \
  --jq '{truncated, manifests: [.tree[]|select(.path|endswith("package.json"))]|length}'
# => {"truncated": false, "manifests": 255}

gh api repos/fufankeji/deepseek-harness-studio/contents/packages/bundle/base/package.json \
  --jq '.content' | base64 -d | jq -c '{name, dsh}'
# => {"name":"@deepseek-ai/dsh-base","dsh":{"bundle":{"patch":"./cordis.patch.yml"}}}

gh api repos/fufankeji/deepseek-harness-studio --jq '.fork'
# => false
```

Behaviour, checked in both directions against the scan loop:

| Tree | Expected |
| --- | --- |
| only `@deepseek-ai/dsh-base` declares a bundle | refused |
| a third-party bundle at the root | accepted |
| `@deepseek-ai/dsh-base` **and** a real plugin bundle | accepted |
| plugin bundle in a subpackage | accepted |
| no bundle anywhere | refused, unchanged message |

Removing `@deepseek-ai/dsh-base` from the set flips the first row to accepted, so the guard is doing the work. The third row is the case that made this a record-then-report rather than a return: a plugin repository may legitimately keep a copy of the harness for testing.

Independent of #1339, which refuses the harness repository by name; this refuses copies of it by package name. Either can land without the other.

### DIFF
--- a/scripts/check-submission.mjs
+++ b/scripts/check-submission.mjs
@@ -32,6 +32,19 @@ const MAX_TREE_PKGS = 40
 // identity rather than by contract.
 const FIRST_PARTY_REPOS = new Set(['deepseek-ai/deepseek-harness'])
 
+// Packages published only by the DSH project. A repository containing one of
+// these under `dsh.bundle` is shipping a copy of the harness: measured on the
+// live topic, five repositories carry `packages/bundle/base/package.json` naming
+// `@deepseek-ai/dsh-base` verbatim, reached about 21 manifests into a ~250
+// manifest tree, so this check finds it inside MAX_TREE_PKGS and accepts them.
+// `fork` is false for all five, so they are source copies that neither an owner
+// check nor a fork check detects.
+const FIRST_PARTY_PACKAGES = new Set([
+  '@deepseek-ai/dsh-base',
+  '@deepseek-ai/dsh-web-app',
+  '@deepseek-ai/dsh-headless',
+])
+
 // Entries submitted before the gate existed are judged by the old rules; only
 // the manifest check applies to them. Set to when the rule change landed.
 const GATE_EFFECTIVE_FROM = process.env.GATE_EFFECTIVE_FROM ?? '2026-08-16T00:00:00Z'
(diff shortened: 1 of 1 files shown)


fkysly:
Merged. The design is the right one: recording the vendored package and continuing the scan, rather than refusing on sight, means a plugin repo that keeps a copy of the harness for testing still passes on its own genuine bundle. That distinction is easy to miss and you got it right.

What I verified before merging:

- **No false positives on the existing list.** Ran the modified checker over a 40-entry sample of currently-listed plugins: 40/40 still pass.
- **The refusal path works** — `deepseek-ai/deepseek-harness` is still correctly refused.

What I could **not** verify: the "five repositories carry `packages/bundle/base/package.json` naming `@deepseek-ai/dsh-base`" measurement. GitHub's code-search API returns `total_count: 0` for every query I tried, including controls that should obviously match, so that is my tooling failing rather than evidence against your number. I spot-checked three large listed monorepos by path and none vendor it. Recording this so nobody later reads the comment as independently confirmed — the guard stands on its own logic either way.

---

已合并。设计思路是对的：记录内置包并继续扫描，而不是一见即拒——这样一个为测试而保留 harness 副本的插件仓库，仍能凭自己真实的 bundle 通过。这个区分很容易被忽略，你处理对了。

合并前我验证了：

- **对现有列表无误伤**：用改后的检查器回测了 40 个已收录条目，40/40 仍然通过。
- **拒绝路径有效**：`deepseek-ai/deepseek-harness` 仍会被正确拒绝。

有一项我**没能**核实：注释里「五个仓库的 `packages/bundle/base/package.json` 直接命名为 `@deepseek-ai/dsh-base`」这个测量。GitHub 代码搜索 API 对我发出的每个查询都返回 `total_count: 0`，连本应命中的对照查询也是，所以这是我的工具不可用，而非否定你的数字。我按路径抽查了三个已收录的大型 monorepo，均未内置。写下来是为了避免后人把这条注释当成已被独立确认——无论如何，这个守卫的逻辑本身是成立的。

PR #1387, opened 2026-08-17
Add springbrand-lab/dsh-plugin-market

La-fe:
Adds [springbrand-lab/dsh-plugin-market](https://github.com/springbrand-lab/dsh-plugin-market) under **Plugin Markets**.

### Why this isn't a duplicate

The category is crowded, so the distinguishing part is stated up front rather than buried. Existing entries install into the profile DSH is currently running. This one enumerates every profile — `web`, `headless`, the launched profile, and each directory under the DSH home — and installs into one that isn't running, with explicit activation timing: the current profile restarts automatically, others apply on their next launch. `Jesse-njx/dsh-plugin-manager` also works per profile, but as a CLI (`dsh pm`) rather than a Settings page.

Two smaller differences: it also runs inside SpringBrand Desktop (delegating package operations to that host's service instead of spawning `dsh plugin`), and the Installed view reads each profile's own `package.json`, so it lists plugins installed outside the catalog too.

If a maintainer still reads this as covered by an existing entry, that's a fair call — I'd rather it be judged on that than on the description.

### Description claims → code

| Claim | Where |
| --- | --- |
| Targets any local profile | `src/profile.ts` `listProfiles`, covered by `tests/profile.spec.ts` |
| Install / update / remove | `src/command.ts` `PluginAction` |
| Install-script warnings | `src/client/Marketplace.tsx` (`runsInstallScripts` badge) |
| Installed view covers non-catalog deps | `src/client/Marketplace.tsx` `installedOnlyRows` |
| Automatic restart on the current profile | `src/index.ts` (`restartRequired`), `src/restart.ts` |

The plugin does **not** hot-mount; the README says so directly, and the description claims a restart rather than hiding it.

### Checklist

- [x] `dsh.bundle` in root `package.json`
- [x] `dsh-plugin` topic added
- [x] Published to npm as `@springbrand/dsh-plugin-marketplace`, so installs are prebuilt and skip `allowBuilds`
- [x] Peer range accepts prerelease cordis (`>=4.0.1-rc.1 <5.0.0-0`), per the note in contributing.md
- [x] READMEs regenerated (+1 line per locale, no other diff)
- [x] `generate-readme.mjs --check`, `build-site.mjs`, `awesome-lint` pass locally


fkysly:
Merged. Checked the one thing that decides whether a `market` entry belongs here at all: it is a real installable plugin, not a directory website. `dsh.bundle` is declared, `src/manager.ts`, `src/command.ts`, `src/profile.ts` carry the logic, tests are present, and it drives `dsh plugin` commands across profiles. We turn down catalogue sites — a plugin that manages plugins is a different thing, and this is that.

---

已合并。对 `market` 分类的条目,决定它是否该被收录的关键只有一条:它是不是真正可安装的插件,而不是一个目录站。这个是——声明了 `dsh.bundle`,逻辑在 `src/manager.ts`、`src/command.ts`、`src/profile.ts`,带测试,实际驱动跨 profile 的 `dsh plugin` 命令。我们会拒绝目录站,但「管理插件的插件」是另一回事,而这正是后者。

PR #1415, opened 2026-08-17
Add dsh-memory-porter (memory)

Shiye-10Pages:
Adds `Shiye-10Pages/dsh-memory-porter` under **memory**.

Brings existing Claude / ChatGPT history into DSH — the gap left by plugins that remember from today onward. Three paths: the `memories.json` inside a Claude export (already distilled, so zero tokens and instant), local Claude Code transcripts with no export needed, and conversations distilled through the model the user already configured in DSH, so no extra API key.

The part I'd point at: every memory must carry verbatim evidence, and that evidence is **checked against the source text by code** rather than taken on the model's word. In a real run over a 7-conversation export, 6 of 25 candidates were rejected for not matching. Evidence found only in the assistant's turns is routed to review instead of stored, since that is a restatement rather than the user's own words.

Checklist:

- [x] `data/plugins/Shiye-10Pages__dsh-memory-porter.yml` added (one file, no conflicts)
- [x] Both READMEs regenerated with `node scripts/generate-readme.mjs`
- [x] `dsh.bundle` declared in `package.json`
- [x] Repo older than 1 day, 18 commits
- [x] `dsh-plugin` topic present
- [x] Published on npm as `dsh-memory-porter`, MIT, zero runtime dependencies


fkysly:
Merged. For anything called a "memory porter" the question is where the data goes, so I traced the whole path: the connectors are local-only (a Claude export zip, or `~/.claude/projects/*.jsonl` transcripts read in place), distillation goes through the host's own configured model via `ctx.llm.stream()` rather than a separate service and key, and storage is append-only JSONL under `~/.dsh/memory-porter/`. Nothing leaves the machine.

The four-gate fidelity system in `gate.ts` — provenance, dedup, contradiction, human review — matches the "verbatim evidence checked against source" claim rather than gesturing at it.

---

已合并。凡是叫「memory porter」的东西,关键问题都是数据流向哪里,所以我把整条链路追了一遍:连接器全部本地(Claude 导出的 zip,或就地读取 `~/.claude/projects/*.jsonl` 记录),蒸馏通过宿主已配置的模型经 `ctx.llm.stream()` 进行,而不是另接服务与密钥,存储是 `~/.dsh/memory-porter/` 下的追加式 JSONL。没有任何内容离开本机。

`gate.ts` 里的四道保真闸——来源、去重、矛盾、人工复核——是对「逐字证据与原文核对」这一说法的实现,而不是泛泛而谈。

PR #1433, opened 2026-08-17
Add dsh-preview and dsh-pilot

Viger1:
Two companion browser plugins for frontend work with a dsh agent.

**[dsh-preview](https://github.com/Viger1/dsh-preview)** — the agent verifies what it built: open a page (local files are served automatically), read the rendered DOM and computed styles, check the console and failed requests, exercise the UI, screenshot for the human. Machine verification is deterministic rather than visual, so a text-only model can assert layout facts. Ships a `frontend-verify` skill.

**[dsh-pilot](https://github.com/Viger1/dsh-pilot)** — the agent operates any page: `pilot_snapshot` returns the page as an accessibility tree whose refs are engine-bound (shadow DOM and same-origin iframes included), so actions cannot be misdirected by snapshot ordering. The origin policy follows the dsh session's own approval stance — a `danger-full-access` session is not re-gated, an `ask` session gets a standard approval card — and the decision is enforced by request interception, so redirects and link-outs cannot drift past the entry gate.

Both are MIT, published to npm (prebuilt, so no `allowBuilds` step), bilingual READMEs, tests and CI.

Checklist per contributing.md:
- One file per plugin under `data/plugins/`, named `Viger1__<repo>.yml`
- `category: tools` (matching the other browser-automation entries)
- Descriptions state what the plugins do, no superlatives; the one containing `: ` is quoted
- Both repos declare `dsh.bundle` in package.json with a `cordis.patch.yml` beside it
- Both carry the `dsh-plugin` topic
- Both are older than a day with well over ten commits
- READMEs regenerated with `node scripts/generate-readme.mjs` and committed


fkysly:
Merged, both. Two separate repositories with their own dependencies and source trees, sharing only family branding — not one plugin split in two.

`dsh-pilot` was the one worth reading closely, since "pilot" could plausibly mean it drives the agent itself. It does not: the tools are `pilot_navigate`, `pilot_snapshot`, `pilot_act`, `pilot_wait`, `pilot_screenshot` and `pilot_close`, and it controls a Playwright browser only, never the session's own control flow.

Two design choices worth naming. It reads the DSH session's existing approval policy rather than inventing a second permission system of its own — a browser driver that ships its own notion of "allowed" is a permission model users have to learn twice and will get wrong once. And it refuses to type into password fields unless `allowPasswordFields: true` is set explicitly. Neither is required of you; both are the right call.

---

已合并，两个条目。这是两个独立仓库，各有自己的依赖与源码树，共用的只是同一套产品命名——不是一个插件拆成两个。

`dsh-pilot` 值得细读，因为「pilot」听起来像是它会驱动 agent 本身。它不会：工具是 `pilot_navigate`、`pilot_snapshot`、`pilot_act`、`pilot_wait`、`pilot_screenshot` 与 `pilot_close`，只控制 Playwright 浏览器，不介入会话自身的控制流。

有两处设计值得点名。它复用 DSH 会话既有的审批策略，而不是自造第二套权限系统——一个自带「什么算被允许」定义的浏览器驱动，等于让用户学两遍权限模型，而且必然会有一次搞错。另外，除非显式设置 `allowPasswordFields: true`，否则它拒绝向密码框输入。这两点都不是硬性要求，但都做对了。

PR #1450, opened 2026-08-17
Add jilian-dsh/dsh-rules-manager (rules manager for dsh)

jilian-dsh:
Adds jilian-dsh/dsh-rules-manager to the plugin list.

- dsh.bundle manifest: yes (patch: ./cordis.patch.yml)
- Repo age: since 2026-08-14, 27 commits
- npm: dsh-rules-manager@1.4.5 published
- Topics: dsh-plugin
- Category: dev

One YAML file per plugin as required; READMEs regenerated via scripts/generate-readme.mjs.

PR #1458, opened 2026-08-17
Add jilian-dsh/dsh-rule-engine (rules execution engine)

jilian-dsh:
Adds jilian-dsh/dsh-rule-engine to the plugin list.

- dsh.bundle manifest: yes (patch: ./cordis.patch.yml)
- Repo age: since 2026-08-15, 22 commits
- npm: dsh-rule-engine@0.3.0 published
- Topics: dsh-plugin
- Category: dev

One YAML file per plugin as required; READMEs regenerated via scripts/generate-readme.mjs.


jilian-dsh:
Hi maintainers — this PR's CI doesn't seem to have triggered either (no check runs, status pending). Could you approve/trigger the workflow run when you get a chance? The plugin is published on npm (dsh-rule-engine@0.3.0) and a GitHub Release (v0.3.0 with .tgz asset) is attached. Thanks!

fkysly:
Merged, with a category change pushed to your branch — no action needed.

**Category `dev` → `security`.** I checked this in source rather than going by the name: `lib/index.js:595` calls `ctx.tools.guard()` and returns `kind: "deny"`. That makes it a permission guard — it decides whether a tool call happens at all — and on this list that is `security`, which is where the approval gates and permission rules live. `dev` is for plugin-development tooling and runtime internals.

Your sibling PR #1450 stays in `dev`, correctly: it has no `ctx.tools.guard` and only reads and writes `AGENTS.md`. The two really are separate codebases with their own bundles, which is the other thing I check when two plugins arrive from one author at once.

The free-zone parsing and the post-write rollback in `version-guard.js` are both real, and there is a test suite behind them.

---

已合并,并已推送一处分类改动到你的分支,你无需操作。

**分类 `dev` → `security`。** 这一条我是查源码而非看名字:`lib/index.js:595` 调用 `ctx.tools.guard()` 并返回 `kind: "deny"`。这使它成为权限守卫——它决定一次工具调用是否发生——而在本列表中这属于 `security`,审批门与权限规则都在那里。`dev` 是给插件开发工具与运行时内部机制用的。

你的另一个 PR #1450 保留在 `dev`,这是对的:它没有 `ctx.tools.guard`,只读写 `AGENTS.md`。两者确实是各自独立的代码库、各有自己的 bundle——这是同一作者一次提交两个插件时我会检查的另一件事。

自由区解析与 `version-guard.js` 中的写后回滚都是真实实现,背后也有测试。

jilian-dsh:
@fkysly Thank you for the review, the merge, and the category fix pushed to our branch — much appreciated. We agree with `security`: the engine does call ctx.tools.guard() and can deny tool calls, so it acts as a permission gate. Happy to adjust anything else if needed. Thanks again!

@fkysly 感谢审核、合并以及推送到我们分支的分类修正——帮了大忙。我们认同归 security：引擎确实调用 ctx.tools.guard() 并能拒绝工具调用，是权限门。如需其他调整请随时告知。再次感谢！

PR #1495, opened 2026-08-17
Add MichengAI/dsh-automation

MichengAI:
Adds [MichengAI/dsh-automation](https://github.com/MichengAI/dsh-automation) under Workflow & Automation.

Eligibility: dsh.bundle, cordis.patch.yml, 1+ day, 18+ commits, dsh-plugin topic, official packages as peerDependencies.

Validation: node scripts/generate-readme.mjs


fkysly:
Reviewed and queued. The safety design is the part worth naming: scheduled runs default to read-only, `workspace-write` is an explicit opt-in, unattended `danger-full-access` is not offered at all, and approval is fail-closed. A scheduler that runs the agent unattended is exactly where a permissive default does real damage, and yours is closed by default and documented as such.

Real test suite too — one spec file per source module.

---

已审完，排队合并。安全设计是最值得点名的部分：定时任务默认只读，`workspace-write` 需显式开启，无人值守的 `danger-full-access` 根本没有提供，审批为 fail-closed。一个让 agent 在无人看管时运行的调度器，正是宽松默认值会造成实际损害的地方，而你这里默认是关闭的，并且写进了文档。

测试也是真的——每个源码模块对应一个 spec 文件。

PR #1508, opened 2026-08-17
Add opencues/opencues#integrations-dsh

wkasekende:
Adds OpenCues for DeepSeek Harness.

**What it does:** word alternatives and underscore-gated fill-ins in the composer. You end a line with `_` and it gets filled; misspellings are flagged as you type. It needs no API key of its own, because it routes through `ctx.llm` — the model dsh is already configured with.

**Install:** `dsh plugin --profile web add @opencues/dsh`

Published to npm with the browser bundle prebuilt, so installs skip the `allowBuilds` build-approval step.

Against the stated requirements:

- **`dsh.bundle` declared** — the manifest declares `bundle` (the `cordis.patch.yml` layer) *and* `client`, not `client` alone.
- **Monorepo subpackage** — `url` points at the subdirectory, `name` is `owner/repo#subname`, filename is `opencues__opencues--integrations-dsh.yml`, per contributing.md.
- **Age / activity** — the repo is long-standing and actively maintained; the dsh integration itself merged today.
- **Real working code** — verified end to end against the *published* package on a clean machine state: fresh `HOME`, no `.cues` config, and every OpenCues provider key removed from the environment. `dsh plugin --profile web add @opencues/dsh` installs in seconds with no build prompt, and `the capital of iceland is _` fills to `Reykjavik` on DeepSeek's own model.
- **`zh` omitted** — happy to add a translation if you would rather I did, rather than leave it to a maintainer.

READMEs regenerated with `npm ci && node scripts/generate-readme.mjs` (1248 entries); the only diff is the one added line in each.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,7 @@ A listing isn't permanent either: entries whose repos go away, stop being mainta
 - [omdsh-dev/ex-setting](https://github.com/omdsh-dev/ex-setting) - Settings extensions for DSH.
 - [omdsh-dev/web-components](https://github.com/omdsh-dev/web-components) - Web Components support.
 - [openAGFS/dsh-agfs](https://github.com/openAGFS/dsh-agfs) - File-browser web app for DSH: React frontend and REST API served by the host webserver, a /dsh-agfs command that opens at the current session workspace, and a browse_files model tool.
+- [opencues/opencues#integrations-dsh](https://github.com/opencues/opencues/tree/master/integrations/dsh) - Word alternatives and underscore-gated fill-ins in the composer. End a line with _ and it is filled; misspellings are flagged as you type. Routes through the model dsh is already configured with, so it needs no API key.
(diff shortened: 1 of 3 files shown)


fkysly:
Merged. Verified the subpackage declares its own `dsh.bundle` and `dsh.client` independently of the monorepo root, which is what decides whether a monorepo entry is installable at all.

Credential handling is the part worth naming: key **names** reach the page and the real secrets are substituted server-side for an allowlisted destination, so a browser-side feature never holds the value. The default route uses the model DSH already has configured, so no API key is needed at all unless you opt into your own provider — and that trade-off is documented rather than left to be discovered.

---

已合并。已确认该子包独立于 monorepo 根目录声明了自己的 `dsh.bundle` 与 `dsh.client`，这正是决定一个 monorepo 条目能否安装的关键。

凭据处理是最值得点名的部分：到达页面的是密钥**名称**，真实密钥在服务端针对白名单目标替换，因此浏览器侧功能始终拿不到密钥值。默认路由使用 DSH 已配置的模型，除非你主动选择自己的服务商，否则完全不需要 API key——而这个取舍是写进文档的，不是留给用户自己发现。

PR #1520, opened 2026-08-17
Add omicverse/dsh-omicos

Starlitnightly:
Adds one entry: `data/plugins/omicverse__dsh-omicos.yml`.

**What it does.** Runs OmicVerse/OmicOS bioinformatics analyses from inside dsh. The plugin gives the agent a persistent Python kernel (scanpy / omicverse / R) whose state carries across turns, a search over the installed skill/agent catalog, direct kernel reads (what is bound, and whether an AnnData is already normalized/log1p/scaled), and an account tab for sign-in and subscription.

- Repo: https://github.com/omicverse/dsh-omicos
- npm: `@omicverse/dsh-omicos` (0.2.1) — one package; installing it also brings omicos-core down, so there is no mid-conversation download on first use
- `dsh.bundle.patch` declared; `dsh-plugin` topic set; GPL-3.0

**On the age gate:** the repository was created 2026-08-16T23:32Z, so it is under the 1-day minimum and the Submission gate will flag it. Commits (28) and the manifest check pass. Happy to have the check re-run after 2026-08-17T23:32Z, or to close and resubmit then — whichever you prefer.


fkysly:
Merged. Checked whether it ships or fetches scientific data, since that is where a bioinformatics integration usually surprises people: neither. `@omicverse/omicos` is a real npm dependency pulling a platform binary at install time, spawned as a child process; the plugin repo itself contains only source, tests and build config.

Worth crediting the candour in your security section: tools run in core with `permission_mode: "full"` because a single-shot tool result leaves no room for mid-flight approval, and you say bridging DSH approval UI is planned and not done. Stating a limitation that plainly is more useful to a reader than a reassuring summary would be.

---

已合并。核查了它是否内置或拉取科学数据集，因为生信类集成最容易在这里让人意外：两者都没有。`@omicverse/omicos` 是真实的 npm 依赖，在安装时拉取平台二进制并作为子进程拉起；插件仓库本身只有源码、测试与构建配置。

你安全章节里的坦率值得肯定：工具在 core 中以 `permission_mode: "full"` 运行，因为一次性返回的工具结果没有中途审批的余地；你也写明「接入 DSH 审批界面属于计划中、尚未完成」。把限制说得这么直白，对读者的价值高于一段让人安心的概述。

PR #1521, opened 2026-08-17
Add SodaMem/dsh-plugin-sodamem (memory)

xlows1206:
Adds [SodaMem/dsh-plugin-sodamem](https://github.com/SodaMem/dsh-plugin-sodamem) under `memory`.

Native Cordis plugin. Recall is contributed on `system-prompt/assemble` and retain runs on `agent/turn-stopping`, so neither costs a tool call and the read path makes no model call. Each recalled fact carries the turn it came from and validity dates, so superseded facts stop being retrieved. It talks HTTP to a SodaMem daemon; the memory engine is not bundled.

- [x] One file at `data/plugins/SodaMem__dsh-plugin-sodamem.yml`
- [x] Ran `node scripts/generate-readme.mjs` and committed both regenerated READMEs
- [x] `package.json` declares `dsh.bundle` (`{"bundle":{"patch":"./cordis.patch.yml"}}`, no `dsh.client` — no browser UI shipped)
- [x] Repo ≥ 1 day old, 10 commits
- [x] `category: memory`
- [x] Description states function, no superlatives
- [x] `dsh-plugin` topic present

Recommended:

- 📦 Published to npm as [`dsh-plugin-sodamem@0.1.0`](https://www.npmjs.com/package/dsh-plugin-sodamem)
- 🔗 All `@deepseek-ai/*` are `peerDependencies`; the only runtime dependency is `sodamem` (a zero-dependency HTTP client)
- 🖼️ No screenshots — the plugin ships no UI panel

Install path was verified against a real tarball rather than assumed: `dsh plugin --profile tui add <tgz>` resolves it and `dsh --dump-config` composes the row. Behaviour is verified against a real `dsh` runtime and a live daemon, not only unit tests.


fkysly:
Merged. "Self-hosted" is the claim worth checking on a memory plugin, and it holds: `src/config.ts` requires an `apiUrl` pointing at a daemon you run yourself, with no hardcoded remote endpoint anywhere. Fact extraction happens on your daemon, not here. `src/recall.ts` and `src/retain.ts` also do what the description says — no tool call and no model call on the read path, hooking `system-prompt/assemble` and `agent/turn-stopping` instead.

---

已合并。记忆类插件最该核的是「自托管」这一声明，而它成立：`src/config.ts` 要求提供指向你自建守护进程的 `apiUrl`，全文没有任何硬编码的远端地址。事实抽取发生在你的守护进程上，而不是这里。`src/recall.ts` 与 `src/retain.ts` 的行为也与描述一致——读取路径上不调用工具、不调用模型，而是挂在 `system-prompt/assemble` 与 `agent/turn-stopping` 上。

PR #1524, opened 2026-08-17
Add Max-Null/dsh-plugin-center

Max-Null:
Plugin center for DeepSeek Harness: installed-plugin metadata (source/version/description/categories), a community market with stars and npm versions, one-click install/update, update detection with Update All, and a What's New startup dialog.

Repo: https://github.com/Max-Null/dsh-plugin-center
npm: @max-null/dsh-plugin-center


fkysly:
Merged. Confirmed which of the two things it is, since we do turn down catalogue websites: this runs inside the DSH Web UI and does real local work — lists installed plugins with provenance, browses catalogues, installs and updates in place, and flags peer-range incompatibilities. A plugin that manages plugins is exactly what `market` covers.

The compatibility flagging is a nice touch, incidentally. Pinned harness versions are becoming a real source of trouble as the harness moves through its release candidates, and surfacing that to a user before they install is more than most managers do.

---

已合并。确认了它属于哪一类——因为我们确实会拒绝目录站：这个运行在 DSH Web 界面内，做的是真实的本地操作——列出已安装插件及其来源、浏览目录、就地安装与更新、并标记 peer 版本区间不兼容。「管理插件的插件」正是 `market` 所覆盖的。

顺带一提，兼容性标记这一点做得不错。随着 harness 在各个 rc 版本间推进，锁死的版本号正在成为实际的麻烦来源，而在用户安装前就把这一点呈现出来，是多数管理器没有做的。

PR #1565, opened 2026-08-18
Add dsh-project-context (turn DSH workspaces into projects)

buhuikongpan:
### Add `buhuikongpan/dsh-project-context`

DSH natively only has *workspaces*, not *projects*. This plugin turns each workspace folder into a **project**: every new session auto-injects a compact **project-workspace convention** via `systemPrompt.context` (read code and build a mental model before acting; empty folder → brand-new project, files live here), with a per-session **Project** toggle in the composer (ON by default).

v0.2.0 reflects the sandbox-policy style: the context block always renders for cwd'd sessions — enabled renders the project convention, disabled renders an explicit *downgrade to a plain workspace* notice instead of vanishing; per-session state lives in the session log (`project-context/mode` events), so it survives replay / export / checkpoint.

- Declares `dsh.bundle` + `cordis.patch.yml` (installable via `dsh plugin add`).
- Real, working code in `lib/` (host injection + state endpoint, client toggle UI) plus unit tests (`node --test`).
- CI workflow configured (syntax check + tests).
- `dsh-plugin` topic added.
- Category: `workflow` (matches the project / per-session-context behaviour).

Requirements: repo age > 1 day ✅ (created 2026-08-18), commits ≥ 10 ✅ (10 commits).

Per contributing.md: added only my own `data/plugins/*.yml`, the `added-dates.json` entry, and regenerated both READMEs.


fkysly:
Merged.

One note for next time: the PR also edited `data/added-dates.json` to insert its own line. That file is site-derived — the build works out when an entry first appeared from git history — so a hand-written value there is either redundant or, if it ever disagrees with history, quietly wrong. The value you added was today's date and therefore harmless, so I left it rather than making you push again. Just leave that file alone in future submissions.

---

已合并。

下次注意一点：这个 PR 还改了 `data/added-dates.json`，往里加了自己的一行。那个文件是站点派生的——构建时会从 git 历史推算条目首次出现的时间——所以手写的值要么是多余的，要么一旦和历史对不上就是悄悄错掉的。你填的是当天日期，无害，所以我没让你再推一次。以后提交时不要动那个文件就行。

PR #1570, opened 2026-08-18
Add DshMarketPlace/dsh-plugins-store

fengshao1227:
A `market` entry: an in-DSH plugin for browsing and installing DSH plugins without leaving the harness.

### What it adds to the harness

- **`/store`** — registered with `ctx.commands.register({ name: "store" })` in `src/index.js`, opens an overlay
- a **Settings → Plugins** tab, injected into the `settings.plugins.tab` slot
- **two agent-callable tools**, `dshmarketplace_search` and `dshmarketplace_install`
- a **bundled skill** at `skills/dsh-plugin-store/SKILL.md`
- an **approval dialog before any install**, listing the risk flags detected for that plugin

Each claim in the description points at something in the source; I checked them one by one before opening this, since the contributing guide asks for that specifically.

### Against the requirements

| | |
| --- | --- |
| `dsh.bundle` in the root `package.json` | `{ "patch": "./cordis.patch.yml" }`, with `cordis.patch.yml` at the repo root |
| Published to npm | [`dshmarketplace-plugin@0.1.4`](https://www.npmjs.com/package/dshmarketplace-plugin) — so installs are prebuilt, no `allowBuilds` step |
| `dsh-plugin` topic | yes, alongside `dsh`, `deepseek-harness`, `ai-agents` |
| Repo age | created 2026-08-17T02:55:11Z |
| Commits | 11 |
| Other | MIT, not archived, actively maintained |

### About the PR itself

One new file — `data/plugins/DshMarketPlace__dsh-plugins-store.yml`. The two README lines are output from `node scripts/generate-readme.mjs`, not hand edits: **+1 entry line per README, 0 removed**, and no existing entry is touched.

The branch was rebuilt on current `main` before pushing. Its previous version was 112 commits behind, which is exactly the case the stale-fork guard is there to catch. Run locally against `main` as of `317df20`'s parent, all with real exit codes: `generate-readme.mjs --check` ✓, `awesome-lint` ✓ (0 errors; 35 warnings, the same 35 `main` already has without this entry), `build-site.mjs` ✓.

Disclosure: I also run dshmarketplace.dev, which is the catalogue this plugin queries. Category or wording — happy to take whatever you prefer, or to have a maintainer just change it.


fkysly:
Merged. Checked the thing that decides whether a `market` entry belongs here: it is a plugin managing plugins, not a catalogue website. `/store`, a Settings tab through the `settings.plugins.tab` slot, and two tools — all real.

Two details worth crediting. Installs are constrained by regex in `src/installer.js` to bare npm specifiers or `github:owner/repo[#subpath]` and dispatched as an argv array, so there is no shell interpolation and no way to point it at an arbitrary source. And the `tools/pre-execute` gate lists `riskFlags` before an install rather than after.

You also disclose that you run dshmarketplace.dev, the catalogue it queries. That is a conflict of interest declared rather than discovered, which is the right way round.

---

已合并。`market` 分类的关键判断已核实:它是管理插件的插件,不是目录站。`/store`、通过 `settings.plugins.tab` 插槽注入的设置页、两个工具,都是真实实现。

两处细节值得肯定。安装目标在 `src/installer.js` 中由正则限定为纯 npm 包名或 `github:owner/repo[#subpath]`,并以 argv 数组方式派发,因此不存在 shell 插值,也无法被指向任意来源。另外 `tools/pre-execute` 闸门是在安装**之前**列出 `riskFlags`,而不是之后。

你还主动说明了自己运营着它所查询的目录站 dshmarketplace.dev。利益关系是主动披露而非被人发现的,顺序是对的。

PR #1606, opened 2026-08-18
Update dsh-remote-workspace tarball to v0.1.0

Hefulalala:
dsh-remote-workspace v0.1.0 released (stage 1: connection pooling + cache + append/offset writes). Updates the `tarball:` field to the new release asset.


fkysly:
Merged. Verified the new tarball URL resolves to a real GitHub Release asset — a version bump that points at an artifact nobody published is the failure mode worth checking on these, and this is not that.

---

已合并。已确认新的 tarball 地址指向真实存在的 GitHub Release 资产——这类版本号变更最该防的失败方式，就是指向一个根本没发布的产物，而这里不是。

PR #1607, opened 2026-08-18
Add lemoncat7/dsh-knowledge

lemoncat7:
Adds lemoncat7/dsh-knowledge to the Memory category.

The plugin provides local and remote knowledge bases with project/session-scoped recall, controlled write-back, and an embedded Web management console. It declares a standard dsh.bundle, has the dsh-plugin topic, and ships a prebuilt v0.6.0 package in GitHub Releases.

Both READMEs were regenerated with the official generator.


fkysly:
Merged. Two things checked because a knowledge base invites them. `src/remote-provider.ts` talks only to a `url` you configure — no hardcoded third-party host — and `src/management-proxy.ts` puts bearer auth plus same-origin and CSRF checks in front of the embedded console, so the write-back path is not open to a page in another tab.

---

已合并。知识库类插件有两点必须查。`src/remote-provider.ts` 只连接你自行配置的 `url`，没有硬编码的第三方主机；`src/management-proxy.ts` 在内嵌控制台前加了 bearer 鉴权与同源、CSRF 校验，因此写回路径不会对其他标签页里的网页开放。

PR #1762, opened 2026-08-18
feat: add dsh-auto-continue-429 by haochi72

haochi72:
# Pull Request Title: feat: add dsh-auto-continue-429 by haochi72

## Plugin Summary

| Field | Value |
|---|---|
| Name | `dsh-auto-continue-429` |
| Owner | `haochi72` |
| GitHub | https://github.com/haochi72/dsh-auto-continue-429 |
| Category | `workflow` (Workflow & Automation) |
| npm package | `dsh-auto-continue-429` (latest 0.2.1) |
| Install command | `dsh plugin --profile web add dsh-auto-continue-429` |

## What it does

A dual-side (host + client) Cordis plugin for the DeepSeek Harness Web UI that
automatically recovers from 429 RATE_LIMIT errors:

1. **Host-side recovery** – listens to `session/event` and intercepts
   `turn/end` events whose `reason.error.code` is `RATE_LIMIT` (429) or
   `QUOTA` (`insufficient_quota` and similar exhausted-quota/balance
   failures, after the built-in `dsh-llm-retry` step-level retries are
   exhausted). After a random 1-2 second backoff, it calls `agent.followup()`
   with a user message `"continue"` to start a new turn. The listener is
   global — it covers every session, not just the one open in the current
   page. Consecutive failures are counted **per session** (`sessionState`
   keyed by session id, cleaned up on `session/disposed`), so a burst of 429s
   across many sessions stops each session independently instead of tripping a
   global limit. When any turn ends with a non-error, non-aborted reason the
   counter for that session resets to zero. The upper limit (default 20,
   range 1-100) is editable by the user and persisted to
   `~/.dsh/auto-continue-429.json`.

2. **Client-side UI** – an on/off sliding switch bar anchored to the chat
   toolbar (next to the model selector), plus a dedicated **"429 自动
   Continue"** tab in the Settings page left nav:
   - green `429` pill + sliding switch showing `开`/`关`; clicking it
     toggles the **quick switch** (`quickOn`) from anywhere in the chat,
   - shows a `retryCount / maxRetries` chip (orange when there have
     already been retries; across multiple sessions it shows the highest
     per-session count, with a hover tooltip explaining the semantics),
   - the Settings tab exposes the **plugin master switch** (`enabled`) —
     when the plugin is disabled, the toolbar switch bar is hidden — and a
     **number input for the custom consecutive-failure upper bound (1-100)**
     with inline success/error toasts and Enter-to-save.

3. **Zero credential access** – never reads API keys or writes outside
   `~/.dsh/auto-continue-429.json`; all host↔client communication goes
   through localhost routes registered on the DSH built-in `webServer`:
   - `GET  /api/auto-continue-429/state` (returns per-session counts too)
   - `POST /api/auto-continue-429/toggle`
   - `POST /api/auto-continue-429/toggle-quick`
   - `POST /api/auto-continue-429/hide-button`
   - `POST /api/auto-continue-429/set-max-retries`

## Why it belongs in the registry

`RATE_LIMIT` (429) and exhausted-quota (`QUOTA`, e.g. `insufficient_quota`)
are two of the most frequent reasons a long turn chain gets truncated
mid-work in DSH. The built-in `dsh-llm-retry` plugin only retries within a
single step; once that budget is spent the user must manually type
`continue` — often 10+ times in a row for very long sessions. This plugin
completely automates that loop with a sane per-session upper-bound guard and
a friendly UI, filling a real gap while being strictly opt-in and fully
inspectable.

## Files in this PR

- `data/plugins/haochi72__dsh-auto-continue-429.yml` – registry entry.

## Verification checklist

- [x] Plugin repository is **public** (https://github.com/haochi72/dsh-auto-continue-429).
- [x] `package.json` includes `dsh.bundle.patch` pointing to `./cordis.patch.yml`
      and a `dsh.client` block with `"platform": "web"`.
- [x] A `README.md` with installation instructions, feature list and architecture
      overview is present in the repo root.
- [x] License file present (`MIT`).
- [x] Plugin has been tested locally against the linked `web` profile; both the
      host-side 429-recovery loop and the client-side toolbar switch + settings
      tab injection behave as described.
- [x] Published on npm as `dsh-auto-continue-429` (v0.2.1) so the listed install
      command works out of the box.


haochi72:
Thanks for the review slot! Updating this PR:

- The plugin repo (haochi72/dsh-auto-continue-429) is now **older than 1 day** and has **16 commits**, so the age/commit-count gate should pass on re-run.
- npm package updated to **v0.2.1** (per-session failure counting, session-disposed cleanup, random 1-2s backoff).
- Entry description refreshed to match the current UI (toolbar switch bar + settings tab).

CI shows \ction_required\ — could you approve the workflow run so the gate re-checks? Thanks!

PR #1795, opened 2026-08-18
Update dsh-llm-cost description (session-wide cost + auto-refresh)

chenyinrusi:
Refresh the dsh-llm-cost entry to reflect its current features: per-turn dollar cost under each message, a session cumulative total with hover breakdown, and the LLM+web price auto-refresh tool (replaces the v0.1.0-era wording).


xxiaoxiong (approved):
Accurate copy refresh for dsh-llm-cost. The new description ("per-turn and session-wide" instead of "per-turn, per-step"; "hover breakdown" for the cumulative total; "auto-refresh" instead of "auto-maintenance") better matches what the plugin actually does. EN/ZH and data YAML all in lockstep. Approved.

PR #1807, opened 2026-08-19
add Neumannzc/dsh-archive-manager (archive manager settings UI)

Neumannzc:
Beautified archive-manager settings section for dsh. npm: @tangzai/dsh-ui-archive-manager


fkysly:
The gate caught something real here and I have pushed the fix — no action needed from you.

Your entry pointed at the repository root, but the root `package.json` declares no `dsh.bundle`. The manifest is at `plugin/package.json` (`@tangzai/dsh-ui-archive-manager`), so the install command the site would have printed —

```
dsh plugin --profile web add github:Neumannzc/dsh-archive-manager
```

— would have resolved to nothing. Anyone following it from the list would have got a plugin that appears to install and then does not exist.

Re-pointed using the convention every monorepo entry on the list uses:

```yaml
url: https://github.com/Neumannzc/dsh-archive-manager/tree/main/plugin
name: Neumannzc/dsh-archive-manager#plugin
```

The file is renamed to match (`Neumannzc__dsh-archive-manager--plugin.yml`); the generator derives the filename from the url and will tell you the expected name if it ever drifts.

Worth knowing why this check exists: until recently the gate accepted a `dsh.bundle` found *anywhere* in the repository tree, while the site built the install command from the entry's url. Those agree only when the root itself declares the bundle. An audit found 48 already-listed entries where they did not — listed, CI-green, and uninstallable from the URL printed beside their name. The check now compares the two, and yours is the first submission it caught before merge rather than after.

Description unchanged. `plugin/` really does register a settings section with workspace-grouped archived sessions and a hover unarchive action.

---

门禁在这里抓到了一个真实问题,我已经推了修复——你无需任何操作。

你的条目指向仓库根,而根 `package.json` 并未声明 `dsh.bundle`。manifest 在 `plugin/package.json`(`@tangzai/dsh-ui-archive-manager`),因此站点本会打印出的安装命令 `dsh plugin --profile web add github:Neumannzc/dsh-archive-manager` **解析不到任何东西**。任何照着列表操作的人,会得到一个「看起来在安装、实则不存在」的插件。

已按列表上所有 monorepo 条目通用的惯例改指(见上方英文部分的 url/name),文件也重命名为匹配的 `Neumannzc__dsh-archive-manager--plugin.yml`;生成器会从 url 推导文件名,一旦不一致它会告诉你期望的名字。

顺带说明这条检查为什么存在:直到不久前,门禁接受在仓库树**任何位置**找到的 `dsh.bundle`,而站点是按条目 url 生成安装命令的。两者只有在「根自己声明了 bundle」时才一致。一次排查发现**已收录的条目里有 48 条不一致**——在列表上、CI 全绿、却无法从名字旁那个 URL 安装。现在这条检查会比对二者,而**你这个是它在合并前拦下的第一例**,而不是合并之后才发现。

描述未改动。`plugin/` 确实注册了一个设置页分区,按工作区分组展示归档会话,并支持悬停取消归档。

PR #1898, opened 2026-08-19
Add dsh-plugin-project-management plugin

Luke-Yong:
Add dsh-plugin-project-management plugin


Luke-Yong:
Node 20 is being deprecated. This workflow is running with Node 24 by default. If you need to temporarily use Node 20, you can set the ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true environment variable. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
Run actions/github-script@v7
Error: gate examined no entries despite changed files

PR #2082, opened 2026-08-20
Add lanbaolu/dsh-wechat-bridge (WeChat bridge)

lanbaolu:
Add [lanbaolu/dsh-wechat-bridge](https://github.com/lanbaolu/dsh-wechat-bridge) under **Notifications & Integrations** (`notify`).

- Category: `notify` (consistent with existing WeChat/iLink/IM bridges on the list)
- Declares `dsh.bundle` in `package.json` (installable via `dsh plugin --profile web add @lanbaolu/dsh-wechat-bridge`)
- npm package: `@lanbaolu/dsh-wechat-bridge` v0.3.0 (`repository` points back at the repo)
- Repo topics include `dsh-plugin`
- Local `check-submission.mjs` against the merge base: `checking 1 entry — all checked entries pass`

One-line description verified against the code: model tools `wechat_bridge_start/status/stop/restart/logs/setup/list_projects/select_project` and `wechat_notify`, slash commands in `src/bridge/commands/`, web panel in `src/client/` (`settings.section` slot).


fkysly:
Reviewed the code rather than the README, because this is the plugin shape where the difference matters: it takes messages off an external chat network and lets them drive an agent running on the submitter's own machine.

**The good news first, since it is the part I expected to be missing.** There *is* a sender check, and ordinary messages go through it:

```ts
// src/bridge/main.ts:420  (handleMessage)
if (account.userId && msg.from_user_id !== account.userId) return;
```

`userId` comes from the iLink login and `login.ts:101` refuses to save an account without it, so in the normal QR-binding flow that comparison is real. Two other bridges are currently held here for having nothing at all in this position, so credit where it is due.

**But one path skips it.** `handlePriorityCommand` runs *before* `handleMessage`, straight off the poll loop:

```ts
// src/bridge/main.ts:373-377
onMessage: async (msg: WeixinMessage) => {
  if (handlePriorityCommand(msg)) return;
  messageQueue.push(msg);
  drainQueue();
},
```

and that function never looks at `msg.from_user_id` (`main.ts:347-370`) — it matches on the text alone:

```ts
if (!/^\/(?:stop|clear|new)(?:\s|$)/i.test(text)) return false;
if (session.state !== 'processing') return false;
messageQueue.length = 0;
```

So while a turn is running, **anyone who can send a message to the bound WeChat account** — any contact — can send `/stop`, `/clear` or `/new` and have it obeyed: the queue is emptied, the in-flight turn is cancelled via `client.stop()`, or the session is cleared outright via `sessionStore.clear()`. It is not arbitrary execution; the agent-facing path is still gated. It is unauthenticated cancellation and state loss on someone's running work, triggered by a stranger typing four characters.

`msg.from_user_id` is already in scope, so the fix is one line at the top of `handlePriorityCommand`:

```ts
if (account?.userId && msg.from_user_id !== account.userId) return false;
```

**One more thing, less urgent.** The guard's shape — `if (account.userId && ...)` — is fail-open: an account record with an empty `userId` accepts every sender. Today `login.ts` will not write one, so this is latent rather than live. I would still invert it, because the safe reading of "we do not know who owns this bridge" is *nobody*, not *everybody*, and that should not depend on a check three files away staying strict.

Everything else checked out: `dsh.bundle` is declared, the iLink client is vendored under `src/bridge/wechat/` rather than fetched at runtime, and traffic goes to `ilinkai.weixin.qq.com` — WeChat's own host, nothing of yours in the middle.

Fix the priority-command path and I will take another look. The listing decision on bridges of this kind as a category is a separate question the maintainer is weighing; this comment is only about the defect.

---

我读的是代码不是 README，因为这类插件恰恰是两者差别最要紧的：它把外部聊天网络来的消息，交给跑在提交者本机上的 agent。

**先说好的部分，因为这正是我以为会缺的东西。** 发送者校验是**有**的，普通消息也确实走了它：

```ts
// src/bridge/main.ts:420  (handleMessage)
if (account.userId && msg.from_user_id !== account.userId) return;
```

`userId` 来自 iLink 登录，而 `login.ts:101` 在拿不到它时会拒绝保存账号，所以正常扫码绑定流程下这个比较是实打实的。目前有另外两个桥接插件正因为这个位置上什么都没有而被搁置，所以这一点该给你记上。

**但有一条路径绕过了它。** `handlePriorityCommand` 在 `handleMessage` **之前**执行，直接挂在轮询回调上：

```ts
// src/bridge/main.ts:373-377
onMessage: async (msg: WeixinMessage) => {
  if (handlePriorityCommand(msg)) return;
  messageQueue.push(msg);
  drainQueue();
},
```

而这个函数从头到尾没有看过 `msg.from_user_id`（`main.ts:347-370`），只匹配文本：

```ts
if (!/^\/(?:stop|clear|new)(?:\s|$)/i.test(text)) return false;
if (session.state !== 'processing') return false;
messageQueue.length = 0;
```

于是在一轮对话进行中时，**任何能给这个绑定微信发消息的人**——任何一个联系人——发 `/stop`、`/clear` 或 `/new` 都会被执行：队列清空、通过 `client.stop()` 中断进行中的这一轮，或者直接 `sessionStore.clear()` 清掉会话。这不是任意执行，通向 agent 的那条路仍然是有门的。但它是对别人正在跑的工作的**无认证中断与状态丢失**，代价是陌生人敲四个字符。

`msg.from_user_id` 就在作用域里，所以修复是 `handlePriorityCommand` 开头的一行：

```ts
if (account?.userId && msg.from_user_id !== account.userId) return false;
```

**还有一处，没那么紧急。** 这个守卫的形状——`if (account.userId && ...)`——是 fail-open：`userId` 为空的账号记录会接受所有发送者。目前 `login.ts` 不会写出这种记录，所以它是潜在的而非现行的。但我仍建议反过来写：「不知道这个桥属于谁」的安全解读是**谁都不行**，而不是**谁都可以**；这一点不该依赖三个文件之外的另一处检查一直保持严格。

其余都没问题：`dsh.bundle` 已声明；iLink 客户端是 vendor 在 `src/bridge/wechat/` 里的，不是运行时去拉的；流量走 `ilinkai.weixin.qq.com`，是微信自己的域名，中间没有你的服务器。

把优先命令那条路径修掉，我再看一轮。至于这一类桥接插件作为一个类别要不要收，是维护者正在权衡的另一个问题；这条评论只针对缺陷本身。

lanbaolu:
Thanks for the thorough review — both points are fixed.

**Priority command path** (`src/bridge/main.ts`): `handlePriorityCommand` now checks the sender up front, exactly as you suggested, before any matching happens:

```ts
const ownerId = account?.userId;
if (!ownerId || msg.from_user_id !== ownerId) return false;
```

So `/stop`, `/clear` and `/new` are now only honored from the bound account owner — a stranger typing four characters can no longer cancel an in-flight turn or wipe the session. Normal messages already went through the same check; now every path does.

**Fail-open → fail-closed** (your less-urgent point): I flipped the guard shape in both `handlePriorityCommand` and `handleMessage` so that an account record with an empty `userId` denies every sender instead of accepting everyone. This is latent today because `login.ts` refuses to save an account without `ilink_user_id`, but the safe reading now matches the code and no longer depends on that check three files away staying strict.

**Verification:** `typecheck` + `build` pass, and the fix is released as **v0.3.1** (npm `latest`, provenance-signed). Commit: `9c8dcbc`.

Happy to take another look at whatever else comes up on the category question separately.

fkysly:
补充一下，把上面那条意见的**性质和范围**说准确——我前面的措辞越界了。

我们不做安全审查。README 顶部和 `contributing.md` 都写着这一点：装任何插件都等于在你机器上跑第三方代码，权限和你本人一样大；收录不代表我们审过它的安全性。你的插件安不安全，不是我们该下的判断。

我们管的是**描述和实际行为对不上**（`contributing.md` 评审第 1 条）。你的条目让读者理解成"我能从聊天软件操作我自己的 agent"，但实际上任何能给机器人发消息的人都能——用的是插件运行者的模型额度、凭据和文件权限。少说的这件事，读者不会预期到。

**所以是二选一，两条都合规：**

1. **加发送者身份判断**（未配置时拒绝所有人，而不是放行所有人）；**或者**
2. **不改代码，把条目描述改成实话**——写明"任何能给机器人发消息的人都能在运行者的机器上发起任务"。

选 2 我照样合，不会因为这个打回。这是你的设计决定，不是我替你做的——我只要求那一行字和代码一致。

---

A correction on the **scope** of the comment above — my earlier wording overstepped.

We don't do security review. The README and `contributing.md` both say so: installing any plugin means running third-party code with your full privileges, and being listed does not mean we vetted it. Whether your plugin is secure isn't our call to make.

What we do check is that the **description matches actual behaviour** (review item 1 in `contributing.md`). Your entry reads as "I can drive my own agent from a chat app", when in fact anyone who can message the bot can — using the operator's model quota, credentials and file permissions. That omission is not something a reader would anticipate.

**So it's a choice of two, and both are acceptable:**

1. **Add a sender identity check** (unset config means nobody, not everybody); **or**
2. **Change nothing in the code and make the entry description say it plainly** — that anyone able to message the bot can start tasks on the operator's machine.

If you pick 2, I'll merge it — you won't be sent back over this. It's your design decision, not mine. All I require is that the line matches the code.

lanbaolu:
Thank you for the correction on scope — appreciated.

We are going with **option 1: a sender identity check**, which is already implemented and released (v0.3.1).

Both message entry points now gate on the bound owner before doing anything:

```ts
// handleMessage (ordinary messages)
if (!account.userId || msg.from_user_id !== account.userId) return;

// handlePriorityCommand (/stop /clear /new)
const ownerId = account?.userId;
if (!ownerId || msg.from_user_id !== ownerId) return false;
```

- `from_user_id` is the WeChat sender id; `account.userId` is the person who QR-scanned and bound the bridge in `login.ts`.
- Fail-closed: an account record with no `userId` denies **everyone**, never allow-all.

So the only party who can start tasks on the operator's machine from the chat app is the binding account owner themselves — the description ("drive your own agent from a chat app") matches the code, and no description change is needed.

This fix ships in v0.3.1 (commit `9c8dcbc`, npm `latest`, CI green). Happy to adjust the line if you still see it as overstating.

PR #2087, opened 2026-08-20
Add dsh-session-nav and dsh-session-archive

kiligzzz:
Add two plugin entries from kiligzzz:

- **dsh-session-nav** (category: ui) — piano-key style in-conversation navigation bar; one key per real user message (built from the FULL on-disk log, not the loaded window), hover preview with the model reply, click-to-jump across full history with auto-paging through the same channel as the official "Load earlier" button.
- **dsh-session-archive** (category: session) — sidebar entry that lists every globally archived session, with read-only preview of the most recent turn, restore to un-archive, and delete with a friendly live-session error.

Both are official dual-face dsh plugins (host + client), published under MIT. `node scripts/generate-readme.mjs` would need to be re-run to refresh the listing.


fkysly:
这个 PR 之前红着，但门禁给你的是一条没用的话——"**这是检查器的 bug，不是你的投稿**"。那句话本身是对的（检查器确实崩了），但它把你挡在了真正的原因外面。

**真正的原因**：`data/plugins/kiligzzz__dsh-session-nav.yml` 里，英文描述含有未加引号的 `": "`：

```yaml
  en: Piano-key style in-conversation navigation bar: one key per real user message...
```

YAML 会把第一个 `": "` 当成键值分隔符，于是整行解析失败。改成加单引号即可：

```yaml
  en: 'Piano-key style in-conversation navigation bar: one key per real user message of the current session. Hover a key to preview that turn, click to jump smoothly; auto-pages older history when the target is outside the virtualized window.'
```

`kiligzzz__dsh-session-archive.yml` 那条没有这个问题，不用动。

**bug 已经修了**（`f53dacbac`）。之前 `readEntries` 抛异常后进程直接死掉、不产出结果，于是工作流只能报"没拿到结果"。现在它会把这条解析错误当作普通的投稿问题报出来，连同上面那段"改成这样"的提示一起贴到 PR 上——所以以后遇到同样问题的人，第一次就能看到怎么改。

你改完那一行推上来，门禁会重新评估你的两个插件本身（年龄、提交数、`dsh.bundle`），然后我来审内容。抱歉让你对着一条误导的报错卡了这么久。

---

This PR has been sitting red with an unhelpful message from the gate — "**this is a bug in the check, not in the submission**". That was true as far as it went (the check did crash), but it kept you from the actual cause.

**The actual cause**: in `data/plugins/kiligzzz__dsh-session-nav.yml`, the English description contains an unquoted `": "`. YAML reads the first `": "` as a key/value separator, so the line fails to parse. Wrapping it in single quotes fixes it — see the block above.

`kiligzzz__dsh-session-archive.yml` doesn't have this problem; leave it alone.

**The bug is fixed** (`f53dacbac`). Previously the parse error escaped and killed the process before it wrote its result, so the workflow could only report that it got nothing back. Now it reports the parse error as the ordinary submission problem it is, and posts that "change it to this" hint straight onto the PR — so the next person to hit this sees the fix immediately.

Push the corrected line and the gate will re-evaluate your two plugins on their own merits (age, commit count, `dsh.bundle`); then I'll review the content. Sorry you were stuck on a misleading error for this long.

PR #2200, opened 2026-08-20
Add huajuan2024/dsh-session-manager

huajuan2024:
## Plugin

- **Repo**: https://github.com/huajuan2024/dsh-session-manager
- **Category**: session
- **Description (en)**: Lists, views, deletes, and exports DeepSeek Harness session histories from a panel under Settings → Plugins; supports JSON or Markdown export of a single session.

## What it does

Adds a panel under **Settings → Plugins** in the DSH Web UI that lets a user browse every known session (title, cwd, updatedAt, turns, steps, agentPreset), open a single session to see its message history, delete non-running sessions, and export any one session as either JSON or Markdown via a browser Blob URL.

The plugin declares the required `dsh.bundle` manifest with `cordis.patch.yml` next to it, so it installs via `dsh plugin add`. File operations are confined to the Host process and `sessionId` is validated against `session-<uuid>` before any read or delete touches the filesystem.

## Checks

- [x] Repository declares `dsh.bundle` in `package.json` with a `cordis.patch.yml` at the repo root
- [x] Repository is older than 1 day and has 10+ commits (10 commits at submission time)
- [x] `dsh-plugin` topic added to the repo
- [x] Description states what the plugin does without superlatives
- [x] YAML regenerated both READMEs without touching other entries


huajuan2024:
The 24h repo-age requirement has now elapsed (current repo age: ~89h). The previous gate run reported `repository is 0.0 days old (needs 1)` — pushing a new commit to re-trigger `pr-check.yml` so the gate can re-evaluate against the current repo age. No description or category change.

PR #2563, opened 2026-08-21
Carry stars/downloads probe data for the moguiyu/dsh-tavily root relist

moguiyu:
## Why

[#2194](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/pull/2194) re-pointed `moguiyu/dsh-tavily` from the monorepo subpackage (`packages/dsh-tool-tavily-search`) to the repo root. `data/stars.json` and `data/downloads.json` are **keyed by entry URL**, so the new root URL had no probe entry — the browser view showed empty stars/downloads for the plugin until the next daily probe refresh.

## Change

Carry the last-known-good probe values from the old subpackage URL to the new root URL, the same repoint convention `data/added-dates.json` documents:

| file | carried value |
|---|---|
| `data/stars.json` | `{ "stars": 3, "checkedAt": "2026-08-19" }` (the subdir probe already inherits the parent repo's stars) |
| `data/downloads.json` | `{ "downloads": 1865, "checkedAt": "2026-08-19" }` |

Only the two keys move; nothing else changes (4-line block swap per file).

## Notes

- No `data/plugins` entries are added or changed — the submission gate reports "nothing to verify" by design.
- The next daily probe will re-derive both values for the root URL and overwrite/confirm them, so this is a stop-gap migration, not a permanent override.

### DIFF
--- a/data/downloads.json
+++ b/data/downloads.json
@@ -1227,10 +1227,6 @@
   "downloads": 1250,
   "checkedAt": "2026-08-19"
  },
- "https://github.com/moguiyu/dsh-tavily/tree/main/packages/dsh-tool-tavily-search": {
-  "downloads": 1865,
-  "checkedAt": "2026-08-19"
- },
  "https://github.com/Mombrane/dsh-subagent-monitor": {
   "downloads": 246,
   "checkedAt": "2026-08-19"
@@ -2493,6 +2489,10 @@
  },
  "https://github.com/zylzyqzz/dsh-mobile-pwa": {
   "downloads": 220,
+  "checkedAt": "2026-08-19"
+ },
+ "https://github.com/moguiyu/dsh-tavily": {
+  "downloads": 1865,
   "checkedAt": "2026-08-19"
  }
 }
--- a/data/stars.json
+++ b/data/stars.json
@@ -3307,10 +3307,6 @@
   "stars": 3,
   "checkedAt": "2026-08-19"
  },
- "https://github.com/moguiyu/dsh-tavily/tree/main/packages/dsh-tool-tavily-search": {
-  "stars": 3,
-  "checkedAt": "2026-08-19"
- },
  "https://github.com/Mombrane/dsh-subagent-monitor": {
   "stars": 16,
   "checkedAt": "2026-08-19"
@@ -5973,6 +5969,10 @@
  },
  "https://github.com/zylzyqzz/dsh-mobile-pwa": {
   "stars": 0,
+  "checkedAt": "2026-08-19"
+ },
+ "https://github.com/moguiyu/dsh-tavily": {
+  "stars": 3,
   "checkedAt": "2026-08-19"
  }
 }


fkysly:
已合并。合并后我去线上核了一遍效果，**结果是一半生效、一半不会生效**，跟你说明一下，免得以为问题解决了。

**stars 这半生效了**：线上 `moguiyu/dsh-tavily` 现在显示 `stars = 5`（比你搬过来的快照值 3 还新，说明夜间探测已经按根路径重新抓过了）。

**downloads 这半不会生效**，原因不在你的迁移，而在探测链路：

`scripts/probe-npm.mjs` 是读**条目 URL 指向那个路径**的 `package.json` 来确定 npm 包名的。条目改到仓库根之后，它读到的是：

```json
{ "name": "dsh-tavily-workspace", "private": true,
  "dsh": { "bundle": { "patch": "./cordis.patch.yml" } } }
```

`dsh-tavily-workspace` 是个 `private: true` 的 workspace 根，**npm 上不存在这个包**，所以 `npm` 解析为 null；而 `probe-downloads.mjs` 只对有 npm 包的条目取数，于是 downloads 恒为 null——你搬进 `data/downloads.json` 的那个 1865 不会被读到。

（顺带一提：真正发布的包 `@moguiyu/dsh-tavily` 在 npm 上是存在的，只是条目指向的根路径解析不到它。）

**好消息是条目本身没问题**：根 `package.json` 有 `dsh.bundle.patch`，所以 `dsh plugin add` 装得上，「能装上」这条是满足的。只是这个仓库以 workspace 根的形态收录，就注定拿不到下载量。

**如果希望显示下载量**，可以考虑把条目 URL 指回那个发布 `@moguiyu/dsh-tavily` 的子目录（本列表里 monorepo 条目普遍是这个写法）。但这是取舍——根路径对读者更直观，下载量只是个附带指标。**你不改我也没意见，只是不想让你以为搬了数据就会显示出来。**

---

Merged. I checked the live result afterwards, and **half of it took effect and half won't** — flagging that so you don't assume it's resolved.

**The stars half worked**: the live site now shows `stars = 5` for `moguiyu/dsh-tavily` (newer than the 3 in the snapshot you moved, so the nightly probe has already re-fetched it at the root path).

**The downloads half won't take effect**, not because of your migration but because of how probing works:

`scripts/probe-npm.mjs` determines the npm package name by reading the `package.json` **at the path the entry points to**. With the entry now at the repository root, that's:

```json
{ "name": "dsh-tavily-workspace", "private": true,
  "dsh": { "bundle": { "patch": "./cordis.patch.yml" } } }
```

`dsh-tavily-workspace` is a `private: true` workspace root and **doesn't exist on npm**, so `npm` resolves to null — and `probe-downloads.mjs` only fetches counts for entries with an npm package. Downloads stay null, and the 1865 you moved into `data/downloads.json` is never read.

(For what it's worth, the actually published package `@moguiyu/dsh-tavily` does exist on npm; the root path just doesn't resolve to it.)

**The good news is the entry itself is fine**: the root `package.json` carries `dsh.bundle.patch`, so `dsh plugin add` works and the installability bar is met. It's simply that listing the repository at its workspace root means download counts can't be attributed.

**If you want downloads to show**, consider pointing the entry URL back at the subdirectory that publishes `@moguiyu/dsh-tavily` (that's the usual pattern for monorepo entries here). But it's a trade-off — the root is friendlier for readers, and downloads are a secondary signal. **No objection if you leave it; I just didn't want you thinking the data move would surface it.**

PR #2649, opened 2026-08-21
Add djasdh/interest-memory to Memory category

djasdh:
## What

Adds [djasdh/interest-memory](https://github.com/djasdh/interest-memory) to the **Memory** category.

## Plugin

Cross-session long-term memory for DeepSeek Harness, backed by a local Go service (single binary, ~50MB RAM, SQLite-vec vector store, verify-augmented claims with citation):

- **Zero-lag recall injection** — on every user message, fetches `/api/v1/{agent}/recall` and injects memory as an owned runtime-context snapshot in the same model step (stable KV-cache prefix, no phantom messages).
- **Session-end ingest** — session transcript pushed to the service once the session ends (owned snapshots filtered out).
- **Tools** — `memory_search` / `memory_logs` / `memory_ingest`.

Install: `dsh plugin --profile web add @djasdh/interest-memory-dsh-bridge`

## Compliance

- ✅ `dsh.bundle` manifest declared at **repository root** (`package.json` → `bridge/dsh/cordis.patch.yml`), so `github:djasdh/interest-memory` installs directly
- ✅ Repo created 2026-08-02, far older than 1 day; 90+ commits
- ✅ Real working code (Go service + JS bridge, tests in `bridge/dsh/lib/lib.test.mjs`)
- ✅ `dsh-plugin` topic set on the repo
- ✅ Description states what it does, no marketing terms


fkysly:
描述我要先肯定一点：**你如实写了 "backed by a local Go service"，没有隐瞒依赖。** 卡住的不是诚实度问题，是装完不能用。

`bridge/dsh/lib/index.js` 是个瘦 REST 客户端——`baseUrl` 默认 `http://127.0.0.1:8899`，靠 `shell.run` + `curl` 访问一个独立运行的本地 Go 服务。而那个 Go 服务**不随 `dsh plugin add` 安装**，需要用户另外跑仓库根 README 给的

```
curl -fsSL .../scripts/install.sh | bash
```

（拉源码编译 + 引导配置 + 可选 systemd），或者去 Releases 下预编译二进制自己启动。

装完插件之后，`memory_search` / `memory_ingest` 全部静默空转——你代码注释自己写的 "a dead service never blocking a turn"。**这个降级设计本身是对的**（服务挂了不该卡住对话），但结果是读者按条目描述装完，什么都不会发生，也不会报错。

**两条路，任一即可：**

1. **（更彻底）让插件自己接管 Go 服务的生命周期**——`apply()` 时检测服务在不在，不在就按**锁定的版本号**自动下载官方 release 二进制并托管启动/停止。本列表里 `Tabbit-Browser/dsh-tabbit` 就是这么做的，已经收录，可以参考。
2. **（最低限度）把条目描述改得更直白**——「仅提供协议桥；需要你先按 README 安装并启动 interest-memory 主服务，否则工具静默无响应」。

选 2 我照样合。

另外一个小建议，不阻塞：条目 URL 现在指向仓库根，但真正可安装的是子目录 `bridge/dsh`（已发布为 `@djasdh/interest-memory-dsh-bridge`）。按 monorepo 惯例改成 `https://github.com/djasdh/interest-memory/tree/main/bridge/dsh` 会更准确。

---

Credit where it's due: **you state "backed by a local Go service" plainly — the dependency isn't hidden.** What's holding this isn't candour; it's that the plugin doesn't work after installation.

`bridge/dsh/lib/index.js` is a thin REST client — `baseUrl` defaults to `http://127.0.0.1:8899`, and it reaches a separately-running local Go service through `shell.run` + `curl`. That Go service **is not installed by `dsh plugin add`**; the user has to separately run what the root README gives them:

```
curl -fsSL .../scripts/install.sh | bash
```

(fetch source, build, bootstrap config, optional systemd), or download a prebuilt binary from Releases and start it themselves.

After installing the plugin, `memory_search` / `memory_ingest` silently do nothing — as your own code comment puts it, "a dead service never blocking a turn". **That degradation design is right** (a dead service shouldn't stall a turn), but the result is that a reader installs from the entry, sees nothing happen, and gets no error.

**Either fix works:**

1. **(More thorough) Have the plugin own the Go service's lifecycle** — on `apply()`, check whether it's running and, if not, download the official release binary **at a pinned version** and manage start/stop. `Tabbit-Browser/dsh-tabbit` in this list does exactly that and is worth copying.
2. **(Minimum) Make the entry blunt about it** — "protocol bridge only; you must install and start the interest-memory service per the README first, or the tools silently return nothing."

Option 2 is fine by me.

One non-blocking note: the entry URL points at the repository root, but the installable unit is the `bridge/dsh` subdirectory (published as `@djasdh/interest-memory-dsh-bridge`). Pointing it at `https://github.com/djasdh/interest-memory/tree/main/bridge/dsh` would follow the monorepo convention and be more accurate.

PR #2747, opened 2026-08-22
Add andyfan1094/dsh-codebase-memory and dsh-minimax-usage-pro

andyfan1094:
Adds two plugins by the same author, one file per entry:

- andyfan1094/dsh-codebase-memory (dev)
- andyfan1094/dsh-minimax-usage-pro (usage)

Checklist:

- [x] One file per plugin at `data/plugins/<owner>__<repo>.yml`
- [x] Ran `node scripts/generate-readme.mjs` and committed the regenerated READMEs
- [x] Each repo's `package.json` declares `dsh.bundle` with `cordis.patch.yml` next to it
- [x] Repos are older than 1 day with 10+ commits
- [x] The `dsh-plugin` topic is set on both repos
- [x] Descriptions state functionality only and were checked against each repo's README and tool list
- [x] Optional `tarball` points at each repo's latest GitHub release `.tgz`

Companion PR: three more plugins from the same author (dsh-feishu, dsh-winrm, dsh-github) were submitted separately to respect the per-PR entry cap.


fkysly:
`dsh-minimax-usage-pro` 那条没问题，可以合——两条路由、双 key 名读取顺序都核对过，你在 README 里主动披露了「该路由不在 `PRIVILEGED_METHODS` 名单里、同源浏览器可读」这个取舍，是主动说明而非隐瞒。

**`dsh-codebase-memory` 卡一下。** 你的 `cordis.patch.yml`：

```yaml
command: !!js >-
  process.env.APPDATA + "/npm/node_modules/codebase-memory-mcp/bin/codebase-memory-mcp.exe"
...
failOnStartupError: true
```

两个问题叠在一起：

1. **`process.env.APPDATA` 是 Windows 独有的**，macOS/Linux 上是 `undefined`，拼出来的路径直接就是错的，没有任何跨平台 fallback。
2. **`failOnStartupError: true`** 意味着找不到这个 exe 时插件直接启动失败——而它要求用户在装这个插件**之前**就已经手动 `npm install -g codebase-memory-mcp`。README 只在"卸载"一节顺带提到 `npm uninstall -g`，正文没有安装步骤。

条目描述是：

> Bridges the Codebase Memory MCP code knowledge graph into DSH and keeps indexed repositories fresh with a debounced filesystem watcher

完全没提「仅支持 Windows」和「需要先全局装一个外部 CLI」。非 Windows 用户装完就是启动失败。

**好消息是这不严重**：`codebase-memory-mcp` 是正经发布在 npm 上的包（我查了 registry，`latest: 0.10.8`，与你锁的版本一致，还给了 Windows AMD64 的 SHA-256），不是没有版本锁的野仓库。所以这是**可修的配置问题，不是资格问题**。

**改法，两条选一：**

1. **改成跨平台解析**——用 `npx codebase-memory-mcp` 或探测 PATH，让 macOS/Linux 也能跑。这是最好的解法。
2. **如果就是只支持 Windows**，那在描述里写明白：「仅支持 Windows；需先 `npm install -g codebase-memory-mcp@0.10.8`」。同时建议把 `failOnStartupError` 改成 `false` 并在日志里给出明确提示——启动失败对用户来说很难排查。

任选一条我就合这一条。`minimax-usage-pro` 那条要不要先单独拆出来合，你说了算。

另外提一句：列表里已有 `jiayan-xu/dsh-codebase-memory` 也桥接同一个上游，但侧重点不同（那个偏语义搜索/调用链查询，你这个偏文件监听自动重建索引），**不算重复，不影响收录**。

---

The `dsh-minimax-usage-pro` entry is fine and can go in — I checked both routes and the dual key-name lookup order, and your README openly discloses the trade-off that the route isn't in `PRIVILEGED_METHODS` and is readable from a same-origin browser. That's disclosure, not concealment.

**`dsh-codebase-memory` needs a fix.** Your `cordis.patch.yml`:

```yaml
command: !!js >-
  process.env.APPDATA + "/npm/node_modules/codebase-memory-mcp/bin/codebase-memory-mcp.exe"
...
failOnStartupError: true
```

Two problems compounding:

1. **`process.env.APPDATA` is Windows-only** — it's `undefined` on macOS/Linux, so the assembled path is simply wrong, with no cross-platform fallback.
2. **`failOnStartupError: true`** means the plugin hard-fails when that exe isn't found — and it expects the user to have run `npm install -g codebase-memory-mcp` **before** installing the plugin. Your README only mentions `npm uninstall -g` in passing under "Remove"; there's no install step in the body.

The entry says:

> Bridges the Codebase Memory MCP code knowledge graph into DSH and keeps indexed repositories fresh with a debounced filesystem watcher

with no mention of "Windows only" or "install an external CLI globally first". A non-Windows user gets a startup failure.

**The good news is that this isn't serious**: `codebase-memory-mcp` is a properly published npm package (I checked the registry — `latest: 0.10.8`, matching the version you pin, with a SHA-256 for the Windows AMD64 build). This is a **fixable configuration problem, not an eligibility one**.

**Either fix works:**

1. **Resolve it cross-platform** — use `npx codebase-memory-mcp` or probe PATH so macOS/Linux work too. This is the better answer.
2. **If it really is Windows-only**, say so in the description: "Windows only; requires `npm install -g codebase-memory-mcp@0.10.8` first." I'd also flip `failOnStartupError` to `false` with a clear log message — a hard startup failure is difficult for users to diagnose.

Either one and I'll merge this entry. Whether to split `minimax-usage-pro` out and land it now is up to you.

One more note: the list already has `jiayan-xu/dsh-codebase-memory` bridging the same upstream, but with a different emphasis (semantic search and call-chain queries there, filesystem watching and index rebuilds here). **Not a duplicate; it doesn't affect listing.**

andyfan1094:
Thanks for the careful review — both issues are fixed in [dsh-codebase-memory v0.2.1](https://github.com/andyfan1094/dsh-codebase-memory/releases/tag/v0.2.1), now referenced by the updated tarball in this PR.

**What changed (option 1 + 2 combined):**
- The MCP bridge now launches through **npx**: `npx -y codebase-memory-mcp@0.10.8`, so nothing has to be pre-installed globally, and it works on macOS/Linux too. On Windows (where spawning `npx` without a shell fails) the config resolves to the Node executable running npm's `npx-cli.js` — verified with a real MCP initialize handshake against the spawned server.
- `failOnStartupError` is now `false`, so a failed MCP startup is logged instead of killing plugin load; the cross-platform watcher half keeps working regardless.
- The registry entry description now states plainly: Windows-only MCP bridge (watcher half is cross-platform), auto-installs its npm CLI via npx on first start. README rows updated to match.

`dsh-minimax-usage-pro` stays as-is in this PR per your note that it's ready. Happy to split it out if you'd rather merge that one first.

PR #2813, opened 2026-08-23
Add SpookySandwich/dsh-plugin-rollout-scout

SpookySandwich:
Adds one entry: `data/plugins/SpookySandwich__dsh-plugin-rollout-scout.yml`, plus the regenerated READMEs. No other entry is touched — the diff is +1 line in each README.

**What it does.** Providers sometimes roll a new conversation model out to a subset of accounts. This launches concurrent throwaway probe conversations in a folder you choose, reads each one's chain-of-thought as it streams, and scores it on how its paragraphs open. Probes that read as the older model are cancelled within a second or two so they stop spending; confident catches are kept and protected from the plugin's own cleanup.

**On the description being accurate.** The "13 hand-labelled chains-of-thought" is literal — they live in [`lib/fixtures.js`](https://github.com/SpookySandwich/dsh-plugin-rollout-scout/blob/master/lib/fixtures.js), and the same file is what [`test/classifier.test.mjs`](https://github.com/SpookySandwich/dsh-plugin-rollout-scout/blob/master/test/classifier.test.mjs) asserts against, so the console demo and CI cannot drift apart. The self-check runs them through the shipped `classify`/`wantsDiscard`, not a copy.

**Requirements**
- `package.json` declares `dsh.bundle` → `./cordis.patch.yml` (present, with `id: rollout-scout`)
- 30 commits, repo created 2026-08-20
- `dsh-plugin` topic set
- No dependencies at all, so no `@deepseek-ai/*` peer-range issue
- MIT, bilingual README, `npm test` runs 124 checks with no framework

Not on npm yet, and installable from source — `lib/client.js` is committed, so no build step is needed at install time. Happy to attach a prebuilt tarball to a Release and add `tarball:` if you'd prefer that.


SpookySandwich:
Small correction to my own PR description: the plugin **is** on npm now — [`dsh-plugin-rollout-scout@1.3.1`](https://www.npmjs.com/package/dsh-plugin-rollout-scout), published after I opened this. So the `tarball:` field I offered isn't needed, and installs get the prebuilt path rather than the `allowBuilds` approval step.

No change to the YAML — as far as I can tell npm mapping is resolved from the repository by CI rather than declared in the entry. Happy to add a field if that's wrong.

`repository.url` in package.json points back here, and `lib/client.js` stays committed, so the `github:` install still works for anyone who wants master instead of the release.

PR #2906, opened 2026-08-23
Add dsh-wenan - extract short-video scripts from Douyin/Bilibili/Kuaishou/YouTube

XGL2019:
Add [dsh-wenan](https://github.com/XGL2019/dsh-wenan) to the Tools & Capabilities category.

A DeepSeek Harness tool plugin that extracts short-video scripts/captions (文案/标题/作者/话题/数据) from Douyin, Bilibili, Kuaishou, and YouTube via a bundled Python extractor.

- installs with `dsh plugin add dsh-wenan`
- declares a `dsh.bundle` manifest
- MIT licensed

### DIFF
--- a/README.md
+++ b/README.md
@@ -846,6 +846,7 @@ A listing isn't permanent either: entries whose repos go away, stop being mainta
 
 ### Tools & Capabilities
 
+- [XGL2019/dsh-wenan](https://github.com/XGL2019/dsh-wenan) - Extract short-video scripts/captions from Douyin/Bilibili/Kuaishou/YouTube.
 - [1na-ko/dsh-hdc-bridge](https://github.com/1na-ko/dsh-hdc-bridge) - HarmonyOS device bridge: hdc screenshot/install/log/crash/UI automation loop with read_image, official-first versioned API knowledge (SDK .d.ts + offline bundled docs), and a DevEco CLI build/sign/lint lane.
 - [6Mikao9/dsh-wsl-workspace](https://github.com/6Mikao9/dsh-wsl-workspace) - Add a WSL workspace from the web GUI without needing to install dsh or related tools again inside WSL. Bash commands and file read/write operations run within the local WSL distribution on the host machine, while Windows files remain accessible.
 - [863683348/dsh-plugin-academic-writing](https://github.com/863683348/dsh-plugin-academic-writing) - Academic writing toolkit for DSH agents: paper outlines, title and abstract skeletons, GB/T 7714 / APA / MLA citations, phrasing QA, and a pre-submission checklist.


fkysly:
`lib/wenan_extract.py` 确实随包提交在仓库里，"bundled" 这个字面意思成立。但脚本本身：

```python
import requests
from playwright.async_api import async_playwright
```

这两个都是要 `pip install` 的第三方包，而**仓库里没有 `requirements.txt`**，`lib/index.js` 只是 `execFile` 调这个脚本，不做任何依赖检测。Playwright 还需要浏览器二进制（`playwright install`），也没人管。

结果是：`dsh plugin add dsh-wenan` **能装上**，但第一次调用 `get_video_script` 工具，在没有手动装过这些的机器上会**直接抛异常**，而且报错是 Node 侧的 spawn 失败，用户看不懂发生了什么。

**另外 README 和代码对不上**：`lib/index.js` 在 `scriptPath` 为空时会 fallback 到内置的 `wenan_extract.py`，但 README「配置」一节让用户把 `scriptPath` 指向 `C:/path/to/dy_wenan_service.py`——**这个文件名在仓库里不存在**，看起来是没同步的旧文档。

**三条建议，第 1 条是必须的：**

1. **README 写清前置条件**：需要 Python 3 + `pip install requests playwright` + `playwright install chromium`。**最好带上版本号**，否则用户装到哪个版本完全不确定。
2. **让失败可诊断**：在工具首次调用时检测这两个包在不在，缺失时返回一句清楚的话（"需要 pip install requests playwright"），而不是让用户面对一个裸的 spawn 异常。
3. **修 README 的配置示例**：把 `dy_wenan_service.py` 改成实际的 `wenan_extract.py`，并说明默认就用内置脚本、不用填 `scriptPath`。

做完第 1 条我就合。2、3 是让你的用户少踩坑，不强求。

（顺带：`package.json` 的 `repository.url` 写的是 `dsh-wenan/dsh-wenan.git`，这个 org 不存在，是个死链接，有空一并改掉。）

---

`lib/wenan_extract.py` really is committed in the repository, so "bundled" is literally true. But the script itself does:

```python
import requests
from playwright.async_api import async_playwright
```

Both are third-party packages requiring `pip install`, and **there's no `requirements.txt` in the repository**. `lib/index.js` just `execFile`s the script with no dependency check. Playwright additionally needs browser binaries (`playwright install`), which nothing handles either.

The result: `dsh plugin add dsh-wenan` **succeeds**, but the first call to `get_video_script` on a machine without those installed **throws** — and the error surfaces as a Node-side spawn failure, which tells the user nothing about what went wrong.

**The README also disagrees with the code**: `lib/index.js` falls back to the bundled `wenan_extract.py` when `scriptPath` is empty, but the README's "Configuration" section tells users to point `scriptPath` at `C:/path/to/dy_wenan_service.py` — **a filename that doesn't exist in the repository**. That looks like stale documentation.

**Three suggestions; the first is required:**

1. **State the prerequisites in the README**: Python 3, `pip install requests playwright`, `playwright install chromium`. **Pin versions if you can**, or users land on whatever happens to be current.
2. **Make the failure diagnosable**: check for those packages on the tool's first call and return a clear message ("requires pip install requests playwright") instead of leaving the user with a bare spawn exception.
3. **Fix the README config example**: change `dy_wenan_service.py` to the actual `wenan_extract.py`, and note that the bundled script is used by default so `scriptPath` needn't be set.

Do #1 and I'll merge. #2 and #3 just save your users grief; they're not required.

(Incidentally, `package.json`'s `repository.url` points at `dsh-wenan/dsh-wenan.git` — that org doesn't exist, so it's a dead link. Worth fixing while you're in there.)

PR #2935, opened 2026-08-23
Add CREAIT-nl/dsh-plugins#gen-limit, #think-level, #research-mode

f-gibellini:
Adds the last three entries from the same monorepo, all subpackages: `gen-limit`, `think-level` and `research-mode`. Third and final PR for that repo — #2925 carries `tool-disclosure`, #2934 carries `hookkit`, `web-fetch` and `web-search-searxng` — split to stay under the 3-entry limit.

**gen-limit** — `model`. Caps how many sessions may generate concurrently on a given provider/model, with a settings card. The count is taken at `llm/stream` from the distinct sessions actually generating, so a session parked waiting on a subagent holds no slot and the child is what counts; a subagent spawn against a full model joins the same FIFO queue instead of failing, because a fan-out of eight against a limit of three is a pacing problem, not an error. Useful when the backend is one self-hosted GPU, or a metered API.

**think-level** — `model`. Stores a thinking level per provider *and* model, so reasoning effort outlives the session it was set in. dsh's dropdown is per session and remembered nowhere, and subagents are never asked at all — they run at the provider default whatever the work. Precedence is: session selection > this table > the adapter's own default, so a model with no row behaves exactly as it did before installing.

**research-mode** — `workflow`. Deep research as an agent mode. dsh ships every piece of a research fan-out but not the loop, and a loop the model authors per call re-earns the same structural mistakes per call — invisibly, since the report still arrives and still reads well. So the loop ships as a reviewed constant (plan → adaptive parallel rounds driven by the gaps researchers declare → cited synthesis → adversarial review) and the model supplies parameters. Design ported from `dsh-deep-research` by omdsh-dev (MIT, acknowledged in the package LICENSE), with four fixes — chief among them that questions the round budget never reached leave the loop by name, into the report's own "what this does not cover" section, rather than being silently dropped.

**Requirements.**
- Each subdirectory's `package.json` declares `dsh.bundle.patch` → its own `cordis.patch.yml`.
- Repo has the `dsh-plugin` topic, is public, and is past the age and commit-count bars.
- The behaviour each description claims is covered by that package's own tests — 47 (gen-limit), 78 (think-level), 149 (research-mode), `node --test test/*.test.js`.

**Checks run locally before pushing.** `scripts/check-submission.mjs` against these three: all pass. `npx awesome-lint` reports nothing on the three new lines (its warnings are identical on pristine `main`). `SKIP_PUBLISH_CHECKS=1 node scripts/build-site.mjs` builds 2019 rows across both locales. The diff is +1 line per entry in each README; no existing entry is touched.

https://claude.ai/code/session_01HwnQmENG8nbcedut6XdZa8


f-gibellini:
Superseded by #3079 — same three entries, rebased onto current `main`, resubmitted from the CREAIT-nl fork. Closing this one.

PR #3223, opened 2026-08-25
docs: 收录 Deepseek-Continuity

linxuhao:
Adds one entry: [linxuhao/Deepseek-Continuity](https://github.com/linxuhao/Deepseek-Continuity) — a DSH plugin that gives an agent local image / voice / music / SFX generation with pinned identity, and refuses degenerate output instead of returning it as success.

Category `vision`, following the existing image-generation entries there.

Checked before opening:

- `package.json` declares `dsh.bundle.patch`, with `cordis.patch.yml` at the repo root
- published to npm as [`dsh-plugin-continuity`](https://www.npmjs.com/package/dsh-plugin-continuity) (backend on PyPI as [`dsh-continuity`](https://pypi.org/project/dsh-continuity/))
- repo age and commit count are over the thresholds; `dsh-plugin` topic is set
- `node scripts/generate-readme.mjs` regenerated both READMEs, `--check` is clean, `build-site.mjs` builds, `check-submission.mjs --base HEAD~1` reports `ok`
- only my own entry is touched

Every claim in the description maps to code in that repo: identity pinning is `create_character` / `create_animal` / `create_object` / `create_actor` reused by `subject_image` / `actor_tts`; the rejection is `verify.py`'s `DegenerateOutput` (image standard deviation and audio RMS floors), raised on the generate paths; idle unload is the `AUDIO_IDLE_UNLOAD_S` / per-request lazy-load ladder; the OpenAI-shaped alternative is `IMAGE_API_SERVER` / `ASR_SERVER`, which also make `continuity-setup` skip that half's weights.


linxuhao:
One thing I'd rather flag than have you discover: this plugin isn't only an image plugin. It spans four generation modalities plus transcription — image (`create_character` / `create_animal` / `create_object` / `subject_image`), voice (`create_actor` / `actor_tts`), music (`generate_music`), procedural SFX (`gen_sfx` / `sfx_presets`), and ASR (`transcribe`) — 21 tools in total, held together by one idea rather than by one medium: an asset is pinned once and reused, and degenerate output is refused.

I filed it under `vision` because that heading reads **Vision & Multimodal** and the existing image-generation entries live there. But **Voice & Audio** would be just as defensible — the voice half is the larger half by measurement (the pitch-drift table in the README is about TTS, not images) — and **Tools & Capabilities** would be the honest catch-all.

So: re-file it wherever it serves a reader best, no argument from me. And if an asset/media-generation category is ever worth splitting out the way `voice` was split out of `tools`, this entry is a candidate for it. Say the word and I'll change the YAML myself rather than making you do it.

PR #3250, opened 2026-08-25
Add NecromanAlbert/dsh-skill-slash-fuzzy

NecromanAlbert:
Resolves unique kebab-case skill substrings in DeepSeek Harness slash tokens.

- Repo: https://github.com/NecromanAlbert/dsh-skill-slash-fuzzy
- Install: `dsh plugin --profile desktop add github:NecromanAlbert/dsh-skill-slash-fuzzy`
- Category: skill
- Host rewrites unique `/segment` tokens before official skill injection; client adds a `/` source so unique segments appear in the skill menu.

The repository was created today, so the age gate may fail until ~24h. Leaving this PR open for resubmission after that.


NecromanAlbert:
Submission gate: repository is 0.0 days old (needs 1). Leaving the PR open; will resubmit after ~24h. README regen only added this plugin line.

PR #3345, opened 2026-08-26
fix: display MemSearch name in plugin market

zc277584121:
## Summary

- Change the MemSearch catalog name from `zilliztech/memsearch#dsh` to `zilliztech/memsearch#MemSearch`.
- Regenerate the English and Chinese READMEs.

## Why

The plugin market derives the visible card title from the suffix after `#`. The current catalog value therefore displays the subdirectory name `dsh` instead of the product name `MemSearch`. The plugin URL remains unchanged, so installation continues to target `plugins/dsh`.

## Validation

- `npx awesome-lint`
- `SKIP_PUBLISH_CHECKS=1 node scripts/build-site.mjs`


zc277584121:
@fkysly This is a small follow-up to #3017 that changes only the displayed catalog name from dsh to MemSearch. The plugin URL and installation target remain unchanged. When you have a moment, could you please take a look? Thanks.

PR #3470, opened 2026-08-27
Add dsh-conversation-timeline entry

bycall:
Adds the dsh-conversation-timeline dock-bar plugin (conversation.input.dock timeline with hover user-prompt tooltips and click-to-jump). npm: dsh-conversation-timeline@0.1.0.


bycall:
Not a blocker: the workflow uses `actions/checkout@v4` / `actions/setup-node@v4`, which target the deprecated Node.js 20 runtime and are now forced onto Node.js 24. Consider bumping to `@v5` to clear the deprecation warning.

PR #3543, opened 2026-08-28
Point zhu1090093659's remaining entries at the renamed repository

fkysly:
Requested by the author themselves in [dsh-market#391](https://github.com/dsh-market/dsh-market/issues/391).

[#3498](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/pull/3498) fixed the two entries that were actually broken. These five still resolve — GitHub's rename redirect covers subpaths — but they carry the old repository name into the card, the detail page and the install command, so they read as a different project from the one the author now publishes.

All five verified against the new path before rewriting:

| entry | new URL |
|---|---|
| `packages/dsh-git-graph` | 200 |
| `packages/dsh-plugin-manager` | 200 |
| `packages/dsh-remote-web-ui` | 200 |
| `packages/dsh-skill-explorer` | 200 |
| `packages/dsh-task-board` | 200 |

Renaming changes each entry's slug, and with it its detail-page URL and the giscus term its comment thread hangs from. All five threads are empty, so nothing is orphaned.

Site builds clean: 2398 rows × 2 locales.

**One thing for the next probe run, not this PR:** the author reports `@linxin666/dsh-web-ui-all` is now deprecated in favour of `@linxin666/dsh-web-all` (both at 0.3.6). The aggregate entry's URL was repointed in #3498, so `probe-npm.mjs` will re-key on the new URL and should land on the new package. Worth a look after the next nightly — `probe-npm.mjs` does not appear to check the `deprecated` field at all, so a deprecated package would be mapped silently if one ever sits at the probed path.


fkysly:
拆成两个 PR 重开——提交闸门把这 5 处重命名当成 5 条新收录了（文件名变了就是新增文件），闸门本身没做错，是我一次塞太多。见下面两个。

PR #2609, opened 2026-08-21
feat: add Volcengine Ark DSH plugins

narcilee7:
## Summary

This PR adds two officially maintained Volcengine Ark DSH plugins:

- `volcengine/ark-cli#ark-plan-api` (`model`): registers Ark Agent Plan, Coding Plan and postpaid model routes in the native DSH model picker.
- `volcengine/ark-cli#ark-managed-agents` (`tools`): adds a Managed Agents settings tab and MCP tools to dispatch long-running agent tasks to Ark cloud Managed Agents.

Both plugins ship prebuilt tarballs via the `dsh-plugins/v0.1.0` release in `volcengine/ark-cli`.

## Validation

- [x] Two YAML files added under `data/plugins/` (one per plugin; this PR adds 2 entries, within the 3-entry limit)
- [x] Ran `npm ci && node scripts/generate-readme.mjs` and committed regenerated `README.md` / `README.zh.md`
- [x] Both repos declare `dsh.bundle` in `package.json` with matching `cordis.patch.yml`
- [x] Source repo `volcengine/ark-cli` is >1 day old and has 10+ commits
- [x] Categories are `model` and `tools`, matching plugin functionality
- [x] Descriptions state what each plugin does, no superlatives
- [x] `dsh-plugin` topic is set on `volcengine/ark-cli`
- [x] Tarballs are hosted on GitHub Release: `ark-plan-api.tgz` and `ark-managed-agents.tgz

PR #3576, opened 2026-08-28
Add ml020/dsh-workbuddy (ui)

ml020:
Reskins the DSH blank-session hero — the brand mark and the workspace picker — with a warm, minimal landing view modeled on [Wordless](https://github.com/Austin-Patrician/Wordless). The native composer is not touched.

**What it changes**

| Slot | Before | After |
| --- | --- | --- |
| `conversation.hero.brand.mark` | default dsh logo | WorkBuddy logo tile + wordmark |
| `conversation.hero.workspace` | default workspace picker | 44px title, mode pills, entry chips, composer card |
| `↳ conversation.hero.workspace.directoryFlow` | *(declared by the shipped picker)* | re-declared by this entry, so "Add workspace…" survives |

Both hero slots are `kind: 'single'`, so the plugin registers through `ctx.slots.inject` to supersede the shipped occupants rather than competing with them.

**What it deliberately does not touch**

`conversation.composer.bar` — the native dsh composer keeps its Lexical editor, attachments, @ references, slash commands, model selection and submit machinery.

**Adding a workspace still works**

This is the part a naive reskin silently breaks. The hero calls `renderSlot('conversation.hero.workspace', …)` with no creation callback, so replacing the picker with a plain list would drop "Add workspace…" with no error and no visible symptom until a user goes looking for it.

The entry avoids that the same way the shipped picker does: `slots.register` declares `children: { 'conversation.hero.workspace.directoryFlow': { kind: 'single', scope: 'root' } }` and the injected face supplies `createWorkspace` (routed to `ctx.workspaces.create`) plus a `hooks.directoryFlow` observable over that hole's occupancy. The add row renders only while the hole has an occupant; picking a folder creates the workspace and selects it; a rejected create raises a retryable dialog instead of failing silently.

**Notes for review**

- Frontend only — no host/backend logic, and no package-owned storage. Workspace creation goes through the `workspaces` service, so the plugin never writes behind the host's back.
- `dsh.bundle` is declared (patch: `./cordis.patch.yml`) alongside the `dsh.client` half; the client half follows the `tsdown.client.ts` preset contract (`cjs` + `__ModuleLoader__` banner/footer, `react`/`cordis`/`dsh-client-*` externalized).
- Design tokens are scoped under `.root` with a `--wb-*` prefix, so they do not leak into dsh's `--dsw-*` theme variables.
- 22 tests pass (19 browser + 3 invariant); 0 type errors, 0 lint errors.
- The picker restores focus to the host chip on close (except when closing *into* the directory flow, where focus belongs to the flow), and keeps the keyboard cursor scrolled into view.
- Light/dark themes follow `prefers-color-scheme`.

**Screenshots**

`docs/preview-light.png`, `docs/preview-dark.png`, `docs/preview-selected.png` — captured from a Vite + Playwright harness that mounts the real `BrandMark` and `WorkspaceMenu` components against host-accurate dsh chrome, not a mockup. Advertised to the storefront via `screenshots.json`.

**Install**

```
dsh plugin add ml020/dsh-workbuddy
```


fkysly:
审过了,**现在装不上**:`main` 指向 `lib/index.js`,但仓库里只有 `src/` + `tsdown.config.ts`,没有 `lib/`,也没有 Release 和 npm 包。用户 `dsh plugin add github:ml020/dsh-workbuddy` 装完启动即失败。/ Reviewed — `main` points at `lib/index.js` but only `src/` exists; no release, no npm. Install succeeds, boot fails.

三选一,做到任一我就重审:发 npm / 发带预编译 tarball 的 Release(条目加 `tarball:` 字段,参考 #2609)/ 把编译好的 `lib/` 提交进仓库。

另外这条 PR 还带着 README 改动——修的时候顺便 `git checkout origin/main -- README.md README.zh.md` 丢掉即可,现在投稿只需要那一个 yml(#3832 之后 README 在 main 上自动生成)。

PR #1554, opened 2026-08-18
fix: point DIT plugin entry to public repository

MC-shark:
## Summary

Update the existing `dsh-plugin-dit` directory entry to its new standalone public source repository: https://github.com/cuboteam/dsh-plugin-dit.

The previous URL pointed to a package directory inside the private `cuboteam/Dit` monorepo and returned 404 for community users. The npm package name, feature set, and ownership are unchanged. This is a source-location migration of the already-listed entry, not a new plugin submission.

## Validation

- Regenerated both READMEs from `data/plugins`
- `node scripts/generate-readme.mjs --check`
- `npx awesome-lint`
- `node scripts/build-site.mjs`
- Standalone repository CI covers bundle tests, both Harness transports, and `npm pack --dry-run`

Note: renaming the data file is required by the repository filename validator. If the submission gate treats it as newly added because of the filename change, please review it as a migration of the existing accepted entry.

PR #3608, opened 2026-08-28
fix: follow two plugins that moved, drop one that was folded in

fkysly:
Three catalog entries pointed at subpaths that 404 while their repositories are alive and actively pushed — so the market listed three plugins that cannot be installed. Checking the repository trees rather than assuming abandonment found that two had simply moved up a level.

### Repointed

| entry | was | now |
|---|---|---|
| `text2future/flowix#dsh-flowix-memory` | `app/flowix-dsh-host/bundles/dsh-flowix-memory` | `dsh-flowix-memory` |
| `Tlyer233/dsh-vscode-review#dsh-review` | `packages/dsh-review` | `dsh-review` |

Both new paths carry the plugin's own `package.json` — `dsh-flowix-memory` still declares the `dsh-plugin` keyword, `dsh-review` still declares its `dsh.bundle` patch and client injects — and both new URLs resolve 200.

### Removed

`Tlyer233/dsh-vscode-review#dsh-review-changes` is gone from the repository entirely, with no successor anywhere in the tree. Its two halves consolidated: the DSH side is now the single `dsh-review` plugin, whose client injects the composer UI that `dsh-review-changes` used to provide, and the editor side became the `dsh-review-vscode` **VS Code extension**, which is not a DSH plugin. Repointing it would have been inventing a plugin that no longer exists.

### Description correction

The surviving `dsh-review` description said "no git". The plugin records into a **shadow git repository** — that is its whole mechanism — and it now covers the panel too. Rewritten from the repo's own README.

### Probe data

`data/stars.json` and `data/screenshots.json` are rekeyed to the new URLs. The values survive the move unchanged: stars are repository-level, and the screenshot URLs point at `docs/images/` at the repo root (verified 200), not under the moved subpath. Leaving the old keys would have blanked the stars and screenshots until the next probe. `screenshots.json` is CRLF in this repo and stayed CRLF — the edit is one key, byte-preserving.


fkysly:
Split into #3609 (the gate misfire that blocked it), #3610 (the two repoints) and the removal PR above.

Two reasons, both worth stating rather than quietly re-pushing:

1. **The stale-fork guard was right to stop this.** Three repoints-plus-a-removal came to −3/+2 README entry lines, and the guard trips at removed > 2. It exists to catch a stale fork silently deleting entries, and I happened to land on its boundary — so this splits rather than widening a guard that is doing its job.
2. **The submission gate was wrong to stop it**, and that is a real bug: it re-ran the new-submission age/commit bar on a plugin *moving inside a repository the catalog already lists*, and rejected the move for `7 commits (needs 10)` because the repository had been force-pushed since. #3609 fixes that, keeping #1554's protection for renames that land on a repository nothing has verified.

PR #3804, opened 2026-08-30
Update yoza10635/dsh-argp description for 1.0.0 (dual-engine)

yoza10635:
dsh-argp 1.0.0 shipped 2026-08-29 (npm + GitHub Release). The listed description predates the dual-engine form and is now half-true: Stage-1 (eager per-atom shrink) makes one small LLM call per turn, so the old "no LLM calls during compaction" claim no longer holds at the top level, only for Stage-2 (reference-graph eviction). This PR updates the entry description to the 1.0.0 form, matching the package description in the repo.

Changes:
- data/plugins/yoza10635__dsh-argp.yml — description.en / description.zh only (url, name, category=session untouched)
- README.md / README.zh.md — regenerated via `node scripts/generate-readme.mjs` (single entry, no drift)

No tarball/npm fields added: the package installs from source (dsh.bundle.patch) and the npm mapping is auto-collected per contributing rules.


fkysly:
合了。这次改描述改得对,值得指出来:

旧描述写的是「压缩阶段不调用 LLM」(`no LLM calls during compaction`),而 `dsh-argp@1.0.0` 的 `lib/` 里能直接搜到 `ctx.llm` / `llm.stream` / `llm.resolveModelInfo`——**LLM 是会被调用的**。新描述改成「LLM 提议、确定性守卫裁决」,并把 `0-LLM` 的说法收窄到只描述引用图延迟剪枝那一段,这才和代码对得上。`verbatim` 守卫与 append-only 日志在 `log-access.js` / `argp-graph-engine.js` 里也确实存在。

主动把一句过强的旧承诺收回来,比加新功能更值得记一笔——描述属实是我们唯一真正执行的内容标准。

---

Merged. Worth calling out what changed: the old text claimed *"no LLM calls during compaction"*, but `ctx.llm`, `llm.stream` and `llm.resolveModelInfo` are all right there in `dsh-argp@1.0.0`'s `lib/`. The new wording — LLM proposes, deterministic guards dispose — narrows the `0-LLM` claim to the lazy reference-graph eviction, which is where it actually holds, and the `verbatim` guards and append-only log check out in `log-access.js` / `argp-graph-engine.js`.

Retracting an over-strong claim about your own plugin is worth more here than a feature would be. Description accuracy is the one content standard this list actually enforces.

PR #502, opened 2026-08-15
Add SongMiao-tech/dsh-prompt-optimizer to UI Enhancements

SongMiao-tech:
Add [dsh-prompt-optimizer](https://github.com/SongMiao-tech/dsh-prompt-optimizer) to the UI Enhancements section.

A prompt-optimizer plugin for DeepSeek Harness: an **✨ Optimize Prompt** button under the composer rewrites your draft into a clearer, more executable prompt through a host LLM call, shows a before/after dialog, and writes the result back with one click.

Compliance with CONTRIBUTING.md:
- Root package.json declares a `dsh.bundle` manifest (`{"patch": "./cordis.patch.yml"}`) — installable via `dsh plugin add`.
- Contains real, working code (host Remote + client UI, built and verified).
- `dsh-plugin` topic added to the repository.
- One-line descriptions added to both README.md and README.zh.md.

Repo: https://github.com/SongMiao-tech/dsh-prompt-optimizer

PR #3822, opened 2026-08-30
feat: add 173787247/dsh-tool-budget

173787247:
## Summary
- Add [`173787247/dsh-tool-budget`](https://github.com/173787247/dsh-tool-budget) (category `security`).
- Single-entry PR to satisfy the submission gate.
- READMEs regenerated via `node scripts/generate-readme.mjs`.

## Test plan
- [ ] CI + Submission gate green

PR #3832, opened 2026-08-30
ci: generate the READMEs on main instead of requiring them in every PR

fkysly:
Fixes the structural reason the PR queue does not drain. Companion to #3789; independent of it.

## The problem, measured today

| | count |
|---|---|
| open PRs (non-draft) | 296 |
| `dirty` — conflicting with main | **218** |
| conflicting over anything a human wrote | **0** |

Every one of those 218 conflicts is in `README.md` / `README.zh.md`. Two unrelated submissions in the same category insert adjacent lines into the same generated block, so merging either one puts the other into conflict. The queue re-conflicts itself faster than it can be merged: today I merged 22 PRs and each merge dirtied the next batch.

A second group is stuck the opposite way — the author committed only their `data/plugins/*.yml`, `pr-check` demanded the regeneration, and the PR sat red waiting on a mechanical step. That is 20+ PRs right now, several already content-approved:

```
#2662 #2671 #2691 #2807 #3014 #3215 #3219 #3224 #3245
#3269 #3325 #3349 #3401 #3425 #3441 #3445 #3657 #3723 #3805 …
```

Same root cause in both directions: **a generated artifact is checked in, and every contributor is asked to regenerate it.**

## The change

**`sync-readme.yml` (new)** — on push to `main` touching `data/plugins/`, regenerate and commit. It terminates because `generate-readme.mjs` is deterministic: the commit re-triggers the workflow, the second run finds nothing to do and exits. Kept as its own workflow with its own `contents: write` rather than relaxing `build-site.yml`, so *"the build never writes to the repository"* stays true where that invariant matters.

**`pr-check.yml`** — two shapes accepted, and which one a contributor produces is no longer something they have to know:

- PR does not touch the READMEs → regenerate in the runner, so `awesome-lint` and the build see a consistent tree. Nothing to verify beyond "the generator runs".
- PR does touch them → must still match. A hand-edited README is caught here rather than silently overwritten by the sync later.

**`pr-check.yml` stale-fork guard** — now counts deleted files under `data/plugins/` instead of removed README lines. This one is not optional: a stale fork carrying only yml would show `+0/-0` and walk straight past the guard written to catch it ([#502](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/issues/502) was 32 entries silently removed). Counting entry files is the better signal anyway — one file per entry, deletion unambiguous rather than inferred from a diff of prose.

**`build-site.mjs`** — added-dates fall back to the entry file's first commit when the README line has no history. Needed in two new places: during `pr-check` the line does not exist in history at all, and after merge its history belongs to a bot commit rather than the author's. README history is still consulted **first**, so nothing already published can move.

## Verification

Not assumed — run:

1. Committed a simulated yml-only submission with no README change. Before: `generate-readme.mjs --check` exits 1, which is exactly what strands those 20 PRs. After: build green, `awesome-lint` 0 errors, and the entry dates correctly to its yml commit (`added: 2026-08-30`).
2. Built `plugins.json` with this branch and with current `main`, then diffed `added` for all **2662** existing entries: **0 changed.**

## What it does not fix

Existing PRs that already carry README commits stay conflicted — git cannot un-write those. But once this is in, the fix for them is trivial and I can say so in one comment: drop the README commit, keep the yml, and the conflict is gone for good. New submissions never acquire the problem.

## Risk

The failure mode is a stale README on `main` if `sync-readme` fails — visible immediately (`generate-readme.mjs --check` is in the post-merge health check I run every round) and fixed by one command. Compare with today, where the same staleness is prevented by keeping 218 PRs unmergeable.

### DIFF
--- a/.github/workflows/pr-check.yml
+++ b/.github/workflows/pr-check.yml
@@ -28,15 +28,22 @@ jobs:
         # (#502 would have silently removed 32 entries). Fail when the PR
         # drops more than 2 existing entries; genuine removals are rare and
         # a maintainer can land them directly.
+        #
+        # Counts entry FILES, not README lines. It used to count README lines,
+        # which stops working the moment a PR is allowed to omit the generated
+        # READMEs — a stale fork carrying only data/plugins/ would have shown
+        # +0/-0 and sailed through the one guard written to catch it.
+        # data/plugins/ is also the better signal: one file per entry, and the
+        # deletion is unambiguous rather than inferred from a diff of prose.
         run: |
           git fetch -q origin ${{ github.event.pull_request.base.sha }}
-          removed=$(git diff ${{ github.event.pull_request.base.sha }}...HEAD -- README.md \
-            | grep -cE '^-- \[.+\]\(https://github\.com/' || true)
-          added=$(git diff ${{ github.event.pull_request.base.sha }}...HEAD -- README.md \
-            | grep -cE '^\+- \[.+\]\(https://github\.com/' || true)
-          echo "entry lines: +$added -$removed"
+          removed=$(git diff --name-status --diff-filter=D ${{ github.event.pull_request.base.sha }}...HEAD -- data/plugins \
+            | grep -c '\.yml$' || true)
+          added=$(git diff --name-status --diff-filter=A ${{ github.event.pull_request.base.sha }}...HEAD -- data/plugins \
+            | grep -c '\.yml$' || true)
+          echo "entry files: +$added -$removed"
           if [ "$removed" -gt 2 ] && [ "$removed" -gt "$added" ]; then
-            echo "::error::This PR removes $removed existing entries — the fork looks stale. Sync it with upstream main (GitHub: 'Sync fork') and re-push; a single plugin submission should be +2/-0."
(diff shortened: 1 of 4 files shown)


fkysly:
**Live evidence, an hour after opening this.**

`173787247` submitted eleven WSL plugins as eleven separate PRs (#3822-#3831, #3848). All eleven are correct: `dsh.bundle` declared, committed plain-JS entry, zero dependencies, accurate one-line descriptions, all green. I reviewed the source and merged them in one sitting.

**Five went in. Six are now `dirty`:** #3824, #3825, #3826, #3828, #3831, #3848.

Not one of those six conflicts over anything the author wrote. Their entry files are six distinct paths under `data/plugins/` that cannot collide with each other. They conflict because all eleven land in the same generated README category block, so each merge invalidates the rest — the same author, the same batch, the same afternoon, blocked by an artifact none of them touched on purpose.

With this branch merged, all eleven would have gone in back to back.

The 218-of-296 number in the description is not a backlog that will drain if someone merges harder. It is the steady state of the current design: I merged 27 PRs today and manufactured six new conflicts doing it.

PR #4377, opened 2026-09-05
Add dsh-ticktick

PerryLink:
Adds PerryLink/dsh-ticktick, a TickTick (Dida365) task bridge: a session-header task panel and curated agent tools over the official TickTick MCP endpoint. Declares the dsh.bundle manifest, carries the dsh-plugin topic, and is published to npm as @perrylink/dsh-ticktick.


PerryLink:
﻿The submission gate reported "The gate checked nothing" on this PR (no verdict reached). The sibling entries #4376 and #4378 use the same format and merged fine, so this looks like a gate/environment hiccup rather than a submission problem. Could you re-run the workflow when convenient? Thanks!

PerryLink:
Correction to my earlier comment, plus the fix.

I wrote that the gate "checked nothing" and that this looked like a gate/environment
problem. That was wrong, and I should correct it rather than let it stand: the check
did reach a verdict, and the verdict was about this entry. From the failing job log:

  data/plugins/PerryLink__dsh-ticktick.yml: invalid YAML - bad indentation of a
  mapping entry
    A description containing ": " must be quoted:
      en: TickTick (Dida365) task bridge: ...

The `en` value contains ": ", so YAML reads the remainder of the line as a nested
mapping and the file does not parse. I confirmed the parser rejects it and that
quoting the value makes both the `en` and `zh` descriptions parse to exactly the
intended strings.

The sibling entries merged because their descriptions happen not to contain that
sequence, not because their format differs: I checked all 36 entries already on
main under data/plugins/PerryLink__*.yml and none of them contains an unquoted
": " in a description.

Fixed by quoting that one value, pushed to the branch. Sorry for the five days of
red checks and for the misdiagnosis.

PR #4409, opened 2026-09-05
plugins: add dsh-search-router

bottledkzk:
Add dsh-search-router to the awesome list.

- Repo: https://github.com/bottledkzk/dsh-search-router
- npm: dsh-search-router
- Category: browser
- Summary: multi-backend web_search router (DeepSeek official / AnySearch / Bing / DuckDuckGo / SearXNG) with live switching, fallback-chain status, automatic fallback to DeepSeek official, and a bilingual settings panel.

### DIFF
--- a/data/plugins/bottledkzk__dsh-search-router.yml
+++ b/data/plugins/bottledkzk__dsh-search-router.yml
@@ -1,0 +1,7 @@
+url: https://github.com/bottledkzk/dsh-search-router
+name: bottledkzk/dsh-search-router
+category: browser
+tarball: https://registry.npmjs.org/dsh-search-router/-/dsh-search-router-0.1.0.tgz
+description:
+  zh: "多后端搜索路由插件：将 DeepSeek 官方 / AnySearch / Bing / DuckDuckGo / SearXNG 统一包装为 web_search 的 router provider，支持即时切换、当前后端与回退链查询、失败自动回退 DeepSeek 官方搜索；设置面板中英文自适应并支持 Key 脱敏显示。"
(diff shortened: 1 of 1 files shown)


fkysly:
Read the plugin — the router itself is fine: five backends behind one `web_search` provider, live switching, fallback chain, masked keys in the settings panel. The entry needs one change before it goes in, and it is the `tarball:` line.

```yaml
tarball: https://registry.npmjs.org/dsh-search-router/-/dsh-search-router-0.1.0.tgz
```

Two problems with that, and the second is the one that matters:

1. [contributing.md](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/blob/main/contributing.md) requires the tarball to be an `https` `.tgz` **on GitHub's own release hosting** — the list will not hand users a download link it cannot vouch for. A registry URL is not that.
2. **You do not need the field at all.** `dsh-search-router@0.1.0` is published on npm, so storefronts already resolve it through the registry and offer a prebuilt install. The `tarball:` field exists for plugins that are *not* on npm. Pointing it at the registry re-states what npm already answers, and pins `0.1.0` forever while your published version moves on.

**The fix: delete the `tarball:` line.** Nothing replaces it. Everything else in the entry stays as written.

There is no GitHub release on the repo right now, so the alternative — cutting a release and pointing at its asset — would be extra work for no gain given the package is already on npm.

Push to this branch and it merges. Marking `held: author`.

bottledkzk:
Closing this PR: I've discontinued `dsh-search-router` and will not be maintaining it further, so please **do not merge** this entry — the repository is being made private and the link would no longer resolve.

Apologies for the noise, and thanks for maintaining the list.

PR #4556, opened 2026-09-07
Add zhangmiao03/cloakpack — secret guard: git pre-push + DSH context firewall

zhangmiao03:
One vault, two gates: the **cloakpack** CLI blocks `git push` when credentials are in the pushed tree and fixes it with one command (AES-256-GCM local vault + stable placeholders + `unpack` restore); **dsh-cloak** (packages/dsh-cloak) is the DSH-side context firewall that replaces credentials in tool results with opaque placeholders before they enter model context, with system-prompt guidance and dry-run mode.

Shared engine: 36 high-precision signature families (incl. Feishu/WeCom/Tencent Cloud for the CN ecosystem), sensitive-key-name rules, custom regex rules, inline `cloakpack:allow` allowlist, baseline workflow (existing findings recorded, new ones blocked), entropy gate. 71 tests incl. a real-git E2E; zero runtime dependencies.

- Repo: https://github.com/zhangmiao03/cloakpack
- Plugin path: packages/dsh-cloak (npm: dsh-cloak)
- Docs: bilingual README, threat model, honest exclusions list


fkysly:
This PR edits `README.md` by hand and carries no `data/plugins/` file — which means that even if I merged it, the entry would disappear on the next push to `main`.

The READMEs are a **generated artifact**. `sync-readme.yml` regenerates both from `data/plugins/*.yml` after every merge and force-commits the result, so a line that exists only in the README is deleted the next time anything else lands. It would look like the entry was merged and then silently removed, which is worse than a rejection.

**What to send instead** — one file, `data/plugins/zhangmiao03__cloakpack.yml`:

```yaml
url: https://github.com/zhangmiao03/cloakpack
name: zhangmiao03/cloakpack
category: security
description:
  en: 'One vault, two gates for secrets: a git pre-push guard (cloakpack CLI) plus the dsh-cloak context firewall — credentials become opaque cloak placeholders before they reach the model or leave the machine; 36 signature families, inline allowlist, baseline workflow, dry-run, /cloak command.'
  zh: '一个保险库、两道闸门：git pre-push 守卫（cloakpack CLI）加上 dsh-cloak 上下文防火墙——凭据在抵达模型或离开本机之前先变成不透明占位符；36 类签名、行内白名单、baseline 流程、dry-run 与 /cloak 命令。'
```

Two things to settle while you are in there:

- **Point the entry at the plugin, not the monorepo root** — I checked: your root `package.json` is `cloakpack-monorepo` and declares no `dsh` field at all, so an entry pointing at the root installs nothing. Use:

  ```yaml
  url: https://github.com/zhangmiao03/cloakpack/tree/main/packages/dsh-cloak
  name: zhangmiao03/cloakpack#dsh-cloak
  ```

  and name the file `zhangmiao03__cloakpack--packages-dsh-cloak.yml`. `cli` and `core` stay out of the list; the entry is for the thing a user installs into the harness.
- **Category** — I have guessed `security`; pick whatever fits, nobody gets sent back over a category.

Drop the README changes entirely; they are regenerated on `main` either way. Reopen this branch with just the yml, or open a fresh PR — whichever is easier.

The plugin is not the problem here, the shape is. Marking `held: author`.

PR #4662, opened 2026-09-08
Add ShadowMiner/hermes-dsh-bridge and ShadowMiner/hermes-web-mcp

newbeebnb:
Two new plugins from ShadowMiner: hermes-dsh-bridge (bidirectional Hermes<->DSH bridge, UI-visible dispatch) and hermes-web-mcp (keyless web MCP server, 9 tools). Both declare dsh.bundle manifest. Category: tools + browser.

### DIFF
--- a/data/plugins/ShadowMiner__hermes-dsh-bridge.yml
+++ b/data/plugins/ShadowMiner__hermes-dsh-bridge.yml
@@ -1,0 +1,6 @@
+url: https://github.com/ShadowMiner/hermes-dsh-bridge
+name: ShadowMiner/hermes-dsh-bridge
+category: tools
+description:
+  en: "Bidirectional bridge between Hermes Agent and DeepSeek Harness: dispatch_task with UI-visible sessions in the DSH sidebar, session mirroring, and 10 DSH-to-Hermes tools."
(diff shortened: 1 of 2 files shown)


fkysly:
Re-verified today — the gate's message from 2026-09-09 still stands, so this needs one edit from you before it can go in.

`ShadowMiner/hermes-dsh-bridge` is a monorepo: the root `package.json` is `hermes-dsh-bridge-monorepo` and declares no `dsh` field at all. The manifest lives one level down, in `packages/dsh-hermes-link/`. As written, the entry points a user at the root, and `dsh plugin --profile web add github:ShadowMiner/hermes-dsh-bridge` installs nothing.

**The fix, in `data/plugins/`:**

```yaml
url: https://github.com/ShadowMiner/hermes-dsh-bridge/tree/main/packages/dsh-hermes-link
name: ShadowMiner/hermes-dsh-bridge#dsh-hermes-link
```

and rename the file to `ShadowMiner__hermes-dsh-bridge--packages-dsh-hermes-link.yml` to match. Running `node scripts/generate-readme.mjs` locally will name the expected file for you if you would rather not construct it by hand.

The second entry in this PR, `ShadowMiner/hermes-web-mcp`, passed and needs nothing.

Marking `held: author` so the queue stops treating this as unexamined. Push to this branch when it is changed — no need to open a new PR.

PR #4667, opened 2026-09-08
add chromoany/dsh-notify-me (Notifications & Integrations)

chromoany:
Add [chromoany/dsh-notify-me](https://github.com/chromoany/dsh-notify-me) to **Notifications & Integrations**.

- Installs with `dsh plugin --profile web add dsh-notify-me` (also on npm, v1.1.0).
- Browser-layer reminders for the DSH web UI: system toast + sound + tab-title marker when the agent needs your input (approval / plan review / question) or a reply finishes in the background; master on/off and notification language (zh / en / follow the interface) from the DSH Settings page.
- Screenshot shipped via `screenshots.json` (notify toast preview).

### DIFF
--- a/data/plugins/chromoany__dsh-notify-me.yml
+++ b/data/plugins/chromoany__dsh-notify-me.yml
@@ -1,0 +1,6 @@
+url: https://github.com/chromoany/dsh-notify-me
+name: chromoany/dsh-notify-me
+category: notify
+description:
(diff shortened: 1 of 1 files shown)


chromoany:
Superseded by #4741, which carries the same entry for the same plugin — closing so there is one open PR per entry.

For the record: this one failed only the repo-age gate (0.5 days at submission — exactly what the gate note said would clear on its own). #4741 was opened after the bar passed and its `Submission gate` is green.

The red `check` on both is the repo-wide build break tracked in #4731 (`build-site.mjs` cannot derive an added date for three entries that landed via a merge commit), not this submission. Fix PR: #4743.

PR #4680, opened 2026-09-08
Add yihe-packs-dsh (data/plugins/ljc6413__yihe-packs-dsh.yml)

ljc6413:
Add [yihe-packs-dsh](https://github.com/ljc6413/yihe-packs-dsh) 鈥?YiHe pkg-dev knowledge packs for DeepSeek Harness.

- One file: `data/plugins/ljc6413__yihe-packs-dsh.yml`
- Repo declares a `dsh.bundle` manifest (`dsh.bundle.patch` -> `cordis.patch.yml`), installable via `dsh plugin add yihe-packs-dsh`
- npm package published: [yihe-packs-dsh@0.1.0](https://www.npmjs.com/package/yihe-packs-dsh) (Apache-2.0)
- Contents: 52 domain packs (JSON) + 90 RFB experience files (language stacks & frontier fields), served through a catalog plugin; executable/installer included.

### DIFF
--- a/data/plugins/ljc6413__yihe-packs-dsh.yml
+++ b/data/plugins/ljc6413__yihe-packs-dsh.yml
@@ -1,0 +1,6 @@
+url: https://github.com/ljc6413/yihe-packs-dsh
+name: ljc6413/yihe-packs-dsh
+category: dev
+description:
+  en: 'YiHe pkg-dev knowledge packs for DeepSeek Harness: 52 domain packs and 90 RFB experience files for coding cognition.'
+  zh: '缁?DeepSeek Harness 瑁呬細杩涘寲鐨勭紪绋嬭鐭ョ煡璇嗗簱锛?2 涓鍩熷寘涓?90 涓?RFB 缁忛獙鏂囦欢锛岃鐩?11 闂ㄨ瑷€鏍堜笌鍓嶆部鏂瑰悜銆?


fkysly:
Diagnosed — and the gate's message was wrong, so sorry for the five days it cost you. It told you to re-run the workflow and that a persistent failure would be a bug in the gate. Re-running could never have worked, and the bug was in the *message*.

**What is actually wrong:** `data/plugins/ljc6413__yihe-packs-dsh.yml` is not valid YAML.

```
$ node -e "require('js-yaml').load(require('fs').readFileSync('data/plugins/ljc6413__yihe-packs-dsh.yml','utf8'))"
YAMLException: deficient indentation (7:1)
```

Line 6, your `zh:` description, is mojibake — UTF-8 text opened by an editor as GBK and saved back:

```yaml
  zh: '缁?DeepSeek Harness 瑁呬細杩涘寲鐨勭紪绋嬭鐭ョ煡璇嗗簱锛?2 涓鍩熷寘涓?90 涓?RFB ...
```

The mangling also ate the **closing `'`**, so the scalar never terminates and the parser runs off the end of the file. The entry is therefore never parsed, never listed, and never checked — which is why the gate reported that it examined nothing.

**The English line is fine.** Only `zh:` needs rewriting, from your original source rather than from what is in the file — the round-trip destroyed bytes and they cannot be recovered cleanly. Decoding the garble as GB18030 gets approximately:

> 为 DeepSeek Harness 装会进化的编程认知知识库：52 个领域包与 90 个 RFB 经验文件，覆盖 11 门语言栈与前沿方向。

Please confirm that against what you meant rather than pasting it back verbatim — the `?` positions in the recovery are lost characters, and I am guessing at a few.

Check before you push:

```bash
node -e "require('js-yaml').load(require('fs').readFileSync('data/plugins/ljc6413__yihe-packs-dsh.yml','utf8'))" && echo OK
```

Make sure your editor writes UTF-8 without a BOM. Nothing is wrong with the plugin — this is the entry file.

I have also fixed the gate (`e08e3f06`): it now names the unparsable file and the parse error instead of sending you to re-run. Marking `held: author` — push to this branch and it goes in.

PR #2662, opened 2026-08-21
Add dsh-ding-sound, dsh-wx-push and dsh-wx-remote plugins

wwweljf:
Add two plugins from wwweljf/dsh-plugins:

- **dsh-ding-sound** (category: voice): turn-end notification sound with built-in meme voices, Settings panel (preview/switch/random), custom audio folder.
- **dsh-wx-push** (category: workflow): auto-send DSH turn results to WeChat via iLink bot direct send with real delivery check.

Install: \dsh plugin --profile web add github:wwweljf/dsh-plugins#path:/plugins/dsh-ding-sound\ (and .../dsh-wx-push)

PR #4682, opened 2026-09-08
Update guhanfei-ai/dsh-grafana and dsh-mindmap descriptions

tay3223:
Refreshes both entry descriptions to match current functionality:

- **dsh-grafana**: conversational operations across multiple named Grafana sources, with approval and version checks for dashboard changes.
- **dsh-mindmap**: reworded as a conversational Markdown mindmap workspace (step-by-step editing, live side-panel visualization, open/manage/export).

yml-only change (2 entry files), no README edits.

### DIFF
--- a/data/plugins/guhanfei-ai__dsh-grafana.yml
+++ b/data/plugins/guhanfei-ai__dsh-grafana.yml
@@ -2,5 +2,5 @@ url: https://github.com/guhanfei-ai/dsh-grafana
 name: guhanfei-ai/dsh-grafana
 category: tools
 description:
-  en: 'Grafana dashboard editor: paste a dashboard URL, edit the JSON through conversation, and push it back over the HTTP API.'
-  zh: 'Grafana 大盘编辑器：粘贴大盘 URL，通过对话修改 JSON，再经 HTTP API 推回 Grafana。'
+  en: 'Conversational Grafana operations: inspect live dashboard data across named sources, then make safe, reviewable dashboard changes with approval and version checks.'
+  zh: '对话式 Grafana 运维：跨多个命名源站查看实时大盘数据，再通过审批与版本校验安全、可审阅地修改大盘。'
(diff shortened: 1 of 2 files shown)


tay3223:
Heads-up for triage: the `check` build failure is unrelated to this PR's content.

The entries introduced by merge commit 5599809cf (#2662, wwweljf/dsh-plugins) can't get a derivable `added-date`: the added-date walks in `build-site.mjs` (`git log --diff-filter=A` and the README `-p` pass) skip merge commits, and those ymls/README lines were introduced by the merge itself — no non-merge commit in history adds them. Every PR based on current `main` hits this; #4683 failed identically a few hours earlier.

The two description updates here are yml-only and regenerate-consistent. Happy to rebase as soon as the derivation has a non-merge commit to read (or the walk expands merges with `-m` / `--full-history`).

tay3223:
Follow-up after rebasing onto current `main` (5b7be0b95): the Build step still fails with the same three undated URLs.

The pin from 9bee1da31 (`Pin added dates for wwweljf entries`) is **gone from `main`** — `data/added-dates.json` currently has 447 keys, none of them wwweljf, even though the file content at 9bee1da31 contains all three. The pin commit is still an ancestor of `main`, so a later merge (likely one of today's 03:13–03:38 UTC batch, whose branch predated the pin) must have rewritten the file from a stale base — the reversal lives inside a merge commit, which is also why the path-filtered commit history shows nothing after 09-07.

So PRs based on current `main` stay red (`#4683`, this one) until the baseline is re-pinned or the derivation walk expands merges (`-m`/`--full-history`). The description updates themselves remain a clean yml-only change.

tay3223:
Fix proposed in #4853 — restoring the three lost wwweljf date pins (exact content from 9bee1da31) makes `build-site.mjs` pass on a branch based on current main (verified locally). Once it merges, this PR and the rest of the red pile should re-run green; I'll rebase if anything else has moved by then.

PR #4689, opened 2026-09-08
Add songer522/dsh-launcher

songer522:
Adds one entry: `data/plugins/songer522__dsh-launcher.yml`.

**[songer522/dsh-launcher](https://github.com/songer522/dsh-launcher)** — a macOS menu bar app for the DSH web server, plus the host plugin it reads.

### What it does

The app manages `dsh web` from the status bar: start, open, restart, stop, with no terminal. The plugin half is what makes that reliable. DSH mints a per-process launch token and the bare origin answers 401, so the app previously recovered the URL by grepping its own log — which only exists for a server the app started itself. Started from a terminal, there was nothing to read and the tab was dead.

The plugin runs inside the harness, where the port and token are simply known, and writes them to `~/.config/dsh-launcher/runtime.json`, removing the file on shutdown. `0600` in a `0700` directory, since that URL carries a credential.

### Notes for review

- `dsh.bundle` is declared in the **root** `package.json`, so `dsh plugin add github:songer522/dsh-launcher` installs the thing the entry URL points at.
- No dependencies and no peer ranges at all, so it installs against any harness version and needs no `allowBuilds` approval.
- No top-level `inject`: naming `webServer`/`connection` there would leave the row pending and turn the boot audit fatal in a profile without a web server. It waits from inside `apply()` instead, so a `headless` profile — and the base-only profile `dsh plugin add` creates on first use — boots normally. Both cases were tested.
- `tarball:` points at a version-free asset, so `latest/download/` will not rot on the next release.
- Category `tools`, matching the closest listed analogue ([YV3507/dsh-webui-launcher](https://github.com/YV3507/dsh-webui-launcher)). This ships no browser UI, so `ui` seemed wrong. Happy to be re-filed.

Every claim in the description was checked against the code, and the descriptor lifecycle (write on bind, correct token, removal on `SIGTERM`) was verified end-to-end against a real `dsh` profile installed from the release tarball.

The repository is under the 1-day age bar as of opening — leaving the PR alone per contributing.md, since the gate re-runs itself.


songer522:
Rebased onto current `main` to re-trigger the gate now that the repository has passed the 1-day age bar.

**Submission gate: pass** — *"All 1 submitted entry passes: `dsh.bundle` declared, repo old enough, enough commits."*

The remaining red `check` is not from this PR. It fails in the "Build (locale parity, date derivation, templates)" step with:

```
no added-date derivable for: .../dsh-ding-sound, .../dsh-wx-push, .../dsh-wx-remote
```

Those are three `wwweljf/dsh-plugins` entries, unrelated to this submission — the log contains no mention of `songer522`. The same step is currently failing on `main` itself (most recent run at the time of writing: 2026-09-10T02:25Z), so it looks pre-existing rather than something this PR introduced.

Worth flagging for the queue generally: `regate.yml` skips any PR whose `check` is not `success` (`if (check && check.conclusion !== 'success') continue`). While that build step is broken on `main`, every PR blocked only on the age bar is also excluded from the automatic re-check, so those verdicts stay red until each author pushes by hand — which is what happened here.

The diff is still one entry file plus the one generated line in each README; no existing entries are touched.

songer522:
Following up on the red `check` here, since I traced where it comes from and there is already a fix in flight.

**It is not this PR.** `scripts/build-site.mjs` derives each entry's added-date with `git log --diff-filter=A --format=%cI -- <file>`. `git log` hides merge diffs by default, and those three `wwweljf/dsh-plugins` entry files first appear in a merge commit (`5599809cf`, two parents), so the query returns empty and the build refuses. The error text points at `fetch-depth: 0`, which is a red herring — it reproduces on a full clone with every commit present.

Reproduced three ways locally, on pristine `main` with this PR's entry **absent**:

| tree | entries | result |
|---|---|---|
| `main`, without my entry | 3431 | fails, same 3 URLs |
| `main` + my entry | 3432 | fails, same 3 URLs |
| `main` + my entry, 3 `wwweljf` entries removed | 3429 | **builds** |

`data/added-dates.json` did carry pins for those three (added in `23b8e185`, then `9bee1da3`), but they are absent from `main` today — `091722ee5`, an unrelated plugin submission, reverted them. That is why the `main` run at 13:02 today was green and the ones after it are not.

**#4786 already fixes this properly**, by falling back to `--diff-merges=first-parent` when the plain query finds nothing. I confirmed by running `build-site.mjs` at both PRs' exact CI merge SHAs: `30937d04` (theirs) builds 3431 rows and exits 0, while `4f1db732` (mine) fails on the same three URLs. Same base commit, same three files present in both trees — the only difference is the patched script.

So there is nothing to change here. I have deliberately **not** cherry-picked that fix into this PR, to avoid two open PRs carrying the same edit to `build-site.mjs`. Once #4786 lands, this check should go green on its next run.

The **Submission gate** on this PR is already green: *"All 1 submitted entry passes: `dsh.bundle` declared, repo old enough, enough commits."*

One thing worth flagging for the queue generally: `regate.yml` skips any PR whose `check` is not `success` (`if (check && check.conclusion !== 'success') continue`). While this build failure persists on `main`, every PR blocked only on the age bar is also excluded from the automatic re-check, so those verdicts stay red until each author pushes by hand.

PR #4708, opened 2026-09-09
Add kvmem/dsh-super-advisor

Di-Chai:
Adds [DSH SuperAdvisor](https://github.com/kvmem/dsh-super-advisor) under **Workflow & Automation**. The main model calls `ask_advisor` for a focused second opinion; users can edit the request and approve each send, then receive text advice from a model configured in DSH.

The listing points at the verified [0.1.7 latest-release tarball](https://github.com/kvmem/dsh-super-advisor/releases/latest/download/dsh-super-advisor.tgz). Documented setups include a tool-capable local main model consulting a cloud advisor, and Flash consulting Pro. Quality and cost improvements are goals, not benchmarked claims. The package declares `dsh.bundle`; the repository has the `dsh-plugin` topic.

## Build failure fixed

The existing site build fails on three `wwweljf/dsh-plugins` entries because their canonical filenames were introduced in merge commit `5599809cf0e63ef2249d923c805c134d6b73e9ca`. The default `git log --diff-filter=A` query does not examine merge diffs, so it returns no addition date even with complete history. See the [failing run](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/actions/runs/34328878203/job/102392490516).

This branch includes the exact date-lookup fix from #4709, in its own cherry-picked commit. Only when the existing lookup finds no date does the build retry with first-parent merge diffs and timestamp-only output. Existing resolved dates and the frozen baseline retain their behavior. No workflow, submission rule, or other plugin entry is changed. #4709 remains available for maintainers who prefer to merge the infrastructure fix separately first.

The complete diff contains two files:

- `data/plugins/kvmem__dsh-super-advisor.yml`: the single plugin listing, 7 added lines.
- `scripts/build-site.mjs`: the necessary shared date-lookup repair, 8 added lines and 1 removed line.

Generated READMEs and site output are excluded from the PR.

## Validation and repository age

- Entry schema, duplicate checks, and README generation pass.
- The full site build passes for all 3,432 entries in both locales, including the new listing and the three previously undated entries.
- All 432 currently listed entries covered by the frozen date baseline keep their published dates.
- Plugin validation is documented in [ACCEPTANCE.md](https://github.com/kvmem/dsh-super-advisor/blob/main/ACCEPTANCE.md): 51 automated tests, actual DSH Loader checks, new installation, and upgrade/browser acceptance. This PR does not change plugin code.
[...]

PR #4709, opened 2026-09-09
Fix added-date lookup for entries introduced in merges

Di-Chai:
Site builds currently stop on the three `wwweljf/dsh-plugins` entries because their canonical filenames were introduced by merge commit `5599809cf0e63ef2249d923c805c134d6b73e9ca`. The entry-date fallback uses `git log --diff-filter=A`, which does not include merge diffs by default, so it finds no addition even with full Git history. This also fails the site-build check in #4708: [CI failure](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/actions/runs/34314817518/job/102348716964).

When the existing lookup finds no date, retry with merge diffs against the first parent and suppress patch output so the parser still receives only timestamps. Successful lookups keep their existing dates, and ordinary entries avoid the extra merge-history traversal. The frozen date baseline and README history lookup retain their current behavior.

Validation:

- Reproduced missing dates for `dsh-ding-sound`, `dsh-wx-push`, and `dsh-wx-remote`; the merge-aware query returns `2026-09-08T23:01:39+08:00` for all three.
- Checked ordinary entry histories, including the new advisor submission; their resolved dates are unchanged.
- `node --check scripts/build-site.mjs` passes.
- `node scripts/generate-readme.mjs` and `SKIP_PUBLISH_CHECKS=1 node scripts/build-site.mjs` pass for all 3,431 entries in both locales. The generated catalog recovers the three affected dates and preserves all 432 currently listed entries covered by the frozen date baseline.

This PR only changes the date lookup in `scripts/build-site.mjs`; it adds or modifies no plugin entries.

The identical fix is now also included in #4708 as a separate cherry-picked commit so that submission can run a successful base build before its normal age recheck. This PR remains the infrastructure-only option for maintainers; neither branch changes the submission rules or other plugin entries.

### DIFF
--- a/scripts/build-site.mjs
+++ b/scripts/build-site.mjs
@@ -238,8 +238,15 @@ if (ordered.some((e) => !dates[e.url])) {
       try {
         // Oldest "added" commit for that path. Not `-1`, which git applies
         // before --reverse and would hand back the newest instead.
-        const out = execSync(`git log --diff-filter=A --format=%cI -- ${JSON.stringify(file)}`,
+        let out = execSync(`git log --diff-filter=A --format=%cI -- ${JSON.stringify(file)}`,
           { encoding: 'utf8' }).trim().split('\n').filter(Boolean)
+        // A canonical filename can first appear in a merge, whose diff the
+        // default log hides. Retry only missing dates so existing dates stay
+        // unchanged and ordinary entries do not pay for merge diff traversal.
+        if (!out.length) {
+          out = execSync(`git log --diff-merges=first-parent --no-patch --diff-filter=A --format=%cI -- ${JSON.stringify(file)}`,
+            { encoding: 'utf8' }).trim().split('\n').filter(Boolean)
+        }
         const iso = out[out.length - 1]
         if (iso) dates[e.url] = new Date(iso).toISOString()
       } catch { /* not committed yet — falls through to the error below */ }


chromoany:
Independent verification from outside this PR, because I spent today on the opposite hypothesis and was wrong — the record should say this form works.

I ran the three states on one full clone (5,512 commits, `main` = `5b7be0b9`), same worktree, only line 241 changed, each run `SKIP_PUBLISH_CHECKS=1 node scripts/build-site.mjs`:

| line 241 | undated entries | build |
| --- | --- | --- |
| pristine `main` | **3** (the `wwweljf` trio) | refuse to build (exit 1) |
| `--diff-merges=first-parent` **without** `--no-patch` (my closed #4796) | **279** | refuse to build |
| `--diff-merges=first-parent --no-patch` — **this PR** | **0** | **completed (exit 0)** |

**`--no-patch` is load-bearing, exactly as this PR has it.** Without it, `--diff-merges` emits patch output, so `out[out.length - 1]` is a diff line rather than a timestamp:

```console
$ git log --diff-merges=first-parent --diff-filter=A --format=%cI -- data/plugins/1010n111__dsh-about.yml
… 98 lines, last non-empty line:  \ No newline at end of file
$ node -e "console.log(new Date('\\\\ No newline at end of file').toISOString())"
RangeError: Invalid time value
```

That throws into the bare `catch {}` at line 245 and the entry stays undated — silently, for 279 entries, which then reads as a second bug on `main`. It is not: pristine `main`'s README pass leaves 279 undated and its yml walk resolves 276 of them, leaving the 3 (`[probe] undated after readme-pass: 279` → `after yml-pass: 3`).

**Nothing else changes.** Taking the date the same way (`out[out.length - 1]`), pristine and this form agree on **3,506 of 3,506** entry files that both walks resolve — **zero dates change**. The only entries newly dated are the merge-introduced ones: per file, 0 lines → `2026-09-08T23:01:39+08:00` (the merge commit's own `%cI`, matching the `2026-09-08T15:01:39Z` the README ledger already publishes). The extra merge lines the flag adds for ordinary files cannot shift a value, because the code reads the oldest line.

So this PR alone takes `main` from exit 1 to a completed build with no data pins — and the pins in #4743 stop being needed the moment it lands. (My #4743 asks for that only as the zero-code-risk alternative; details and the retraction of my earlier 279-entry claim are in #4731.)

fkysly:
谢谢——这条是专门为这个 bug 开的修复 PR，写法被采纳了。

**经过**：我先自己推了一版 `--first-parent --diff-filter=A`（`417289cfe`），并在 commit message 里断言「加宽兜底不会移动已发布的日期」。**那句是错的**：`--first-parent` 会把遍历限制在主线，于是**任何添加提交落在侧分支上的条目都会被记成合并日而不是真实日期**——实测 170 个条目的日期被挪后，而且以那个状态上线了约二十分钟。

你们的写法避开了这个坑，而且是**只在默认查询一无所获时才重试**：

```
```

它不限制遍历哪些提交，只改合并的 diff 计算方式；加上那个「找不到才重试」的门，已经有答案的条目就保持原值，3400 多个不需要合并遍历的条目也不用为此付代价。**这个门正是我漏掉的部分。** 已按你们的版本替换（`eb665bc93`），170 个日期恢复。

根因记录在案：9-08 合 [#2662](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/pull/2662) 时我把三个条目文件改名以匹配各自的 url，并在同一个合并提交里重新生成了 README 行——于是文件和 README 行在两个父提交里都不存在，两条账本同时失明，构建按设计拒绝运行（盖"现在"会让日期每次构建都跳变）。主干从那一刻红到今天，站点冻结在 3408 条，130 个与此无关的 PR 在一个作者管不到的步骤上变红。

---

**What happened**: I first pushed my own version, `--first-parent --diff-filter=A` (`417289cfe`), with a commit message claiming that widening the fallback could not move a published date. **That claim was wrong.** `--first-parent` restricts traversal to the mainline, so **any entry whose addition commit lives on a side branch gets the merge date instead of its real one** — 170 entries shifted, and the site was published that way for about twenty minutes.

Your form avoids that, and crucially it retries **only when the default lookup found nothing**. It does not restrict which commits are walked, only how a merge is diffed; with that gate, an entry that already has an answer keeps it, and the 3,400 entries that need no merge traversal do not pay for it. **That gate is the part I missed.** Your version is now on main (`eb665bc93`) and the 170 dates are restored.

Root cause for the record: merging [#2662](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/pull/2662) on Sep 8 renamed three entry files to match their urls and regenerated their README lines in the same merge commit, so neither the files nor the lines exist in either parent. Both ledgers went blind and the build refused to run — by design, since stamping "now" would make dates flap between builds. main was red from that moment, the site froze at 3,408 entries, and 130 unrelated pull requests went red on a step no author controls.

fkysly:

```
```

---

PR #4722, opened 2026-09-09
Rename winditer/dsh-elf to winditer/dsh-temp-chat

winditer:
Rename the plugin listing for **winditer/dsh-elf → winditer/dsh-temp-chat**.

The repository `winditer/dsh-elf` was renamed to `winditer/dsh-temp-chat` (the npm package `dsh-elf` is likewise now `dsh-temp-chat`). This PR updates the existing listing entry to point at the new name:

- `data/plugins/winditer__dsh-elf.yml` → `data/plugins/winditer__dsh-temp-chat.yml`, `url`/`name` updated to `winditer/dsh-temp-chat`
- `data/screenshots.json` keys/URLs updated to `winditer/dsh-temp-chat`
- README.md / README.zh.md regenerated via `node scripts/generate-readme.mjs --check` (in sync)

The repo still declares `dsh.bundle` (no functional change), so it remains installable via `dsh plugin add` under its new package name `dsh-temp-chat`.

No new entry — this only renames the existing one, so no age/commit-count gate applies.

PR #4727, opened 2026-09-09
Add Mlte0907/dsh-teams-x and Mlte0907/dsh-remote-x

Mlte0907:
Two entries, both mine. One PR (2 entries, under the 3-entry limit).

| entry | category | what it does |
| --- | --- | --- |
| `Mlte0907/dsh-teams-x` | `workflow` | Turns one DSH session into a durable multi-agent team: a captain spawns member subagents, drives a dependency-aware task DAG with quality gates (objective/acceptance/inScope/verify), and coordinates through persistent mailboxes; a live activity panel shows the team. |
| `Mlte0907/dsh-remote-x` | `remote` | Remote control and mobile overlay: LAN and public (cloudflared) tunnels with QR-code access, a phone-first UI under 768px, and a LAN reverse-proxy CLI. |

Both repositories:

- declare `dsh.bundle` in `package.json`, with `cordis.patch.yml` alongside;
- ship real, working code (not placeholder or README-only);
- are past the 1-day age bar — created 2026-09-04 (`dsh-remote-x`) and 2026-09-06 (`dsh-teams-x`);
- carry the `dsh-plugin` topic.

Only the two entry files are added. The generated READMEs are untouched, as the contribution guide asks.

Category notes: `dsh-teams-x` orchestrates tasks across subagents, so I filed it under `workflow`; `agi` was the other candidate. `dsh-remote-x` is remote access plus a mobile overlay, which is what `remote` is for. Happy to move either if a better fit exists.

### DIFF
--- a/data/plugins/Mlte0907__dsh-remote-x.yml
+++ b/data/plugins/Mlte0907__dsh-remote-x.yml
@@ -1,0 +1,6 @@
+url: https://github.com/Mlte0907/dsh-remote-x
+name: Mlte0907/dsh-remote-x
+category: remote
+description:
+  en: 'Remote control and mobile overlay for DeepSeek Harness: LAN and public (cloudflared) tunnels with QR-code access, a phone-first mobile UI under 768px, and a LAN reverse-proxy CLI.'
+  zh: 'DeepSeek Harness 的远程控制与移动端覆盖层：局域网与公网（cloudflared）隧道 + 二维码接入、768px 以下的手机端界面，附局域网反向代理 CLI。'
--- a/data/plugins/Mlte0907__dsh-teams-x.yml
+++ b/data/plugins/Mlte0907__dsh-teams-x.yml
@@ -1,0 +1,6 @@
+url: https://github.com/Mlte0907/dsh-teams-x
+name: Mlte0907/dsh-teams-x
+category: workflow
+description:
+  en: 'Orchestrates durable multi-agent teams from one DSH session: a captain spawns member subagents, drives a dependency-aware task DAG with quality gates, and coordinates through persistent mailboxes, with a live activity panel.'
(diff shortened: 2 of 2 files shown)


Mlte0907:
Note on the red `check` job, so it isn't read as a problem with these two entries.

`check` fails in `Build (locale parity, date derivation, templates)` with:

```
```

That is pre-existing on `main` and names three entries that are not in this PR:

- **`Submission gate` on this PR is green** — *"All 2 submitted entries pass: `dsh.bundle` declared, repo old enough, enough commits."*
- The same three URLs fail **every open PR** (#4726, #4724, #4722, …) and **`main`'s own `build-site.yml` runs** — e.g. runs `34323902799` (2026-09-09 07:27Z) and `34307896961` (03:38Z) fail with the identical message.
- **Cause:** those three entry files were introduced by merge commit `5599809c` itself (PR #2662, 2026-09-08). Both of its parents lack them — `data/plugins/wwweljf__dsh-plugins--plugins-dsh-ding-sound.yml` returns 404 at `57de4187` and at `34cb8838`, and is present only from `5599809c` onward. Both date sources in `build-site.mjs` are `git log` walks (`-p -- README.md` at line 212 and `--diff-filter=A -- <path>` at line 241), and `git log` produces no diff for merge commits by default, so neither sees the addition. That is also why only these three entries are affected while the many other merge-commit PRs are fine: their additions live in the branch's own non-merge commits. The script's error text ("reachable only from a shallow clone or a genuinely uncommitted entry") doesn't cover this case.

So nothing here needs changing on my side; a re-run should go green once `main` is fixed (e.g. pin the three URLs in `data/added-dates.json`, or let the derivation see merge commits). I've deliberately not touched those entries — a submission shouldn't rewrite someone else's row.

PR #4732, opened 2026-09-09
Add yindf/taskfold (session)

yindf:
Adds [yindf/taskfold](https://github.com/yindf/taskfold) to **Sessions & Messages**.

**What it does.** Wraps work in named tasks (`task_begin("…")` / `task_end("…")`). When a task closes, its whole span of messages is folded into one titled summary at the next step boundary, and `fold_recall({ fold: N })` regenerates the original content from an on-disk archive at any time.

**Category.** `session` — it operates on the session transcript (message spans), not on a UI surface, a model route, or a tool capability.

**Requirements.**
- `package.json` declares `dsh.bundle.patch: ./cordis.patch.yml` at the repo root.
- Real, working code: 9 offline test files, `npm test` green.
- Repo created 2026-09-02 (older than the 1-day floor), not archived.
- `dsh-plugin` topic present.
- Prebuilt tarball attached to the GitHub Release: `dsh-taskfold-0.31.1.tgz` (the `tarball:` field above).

**Accuracy note.** The summary is written once, while the original span is still in context, so it is not a summary of a summary. Archival is queued on close and drains at the next step boundary — that is why the description says "at the next step boundary" rather than "immediately". Nothing is discarded: the original messages are written to a JSONL archive that `fold_recall` reads back verbatim.

### DIFF
--- a/data/plugins/yindf__taskfold.yml
+++ b/data/plugins/yindf__taskfold.yml
@@ -1,0 +1,7 @@
+url: https://github.com/yindf/taskfold
+name: yindf/taskfold
+category: session
+tarball: https://github.com/yindf/taskfold/releases/download/v0.31.1/dsh-taskfold-0.31.1.tgz
+description:
+  en: 'Wraps work in named tasks and, once a task closes, folds its whole span of messages into one summary at the next step boundary; the original content can be restored on demand.'
+  zh: '用命名任务包住工作过程；任务结束后，在下一个步骤边界把整段消息折叠成一段摘要，原文可随时还原。'


yindf:
Heads-up: the failing `check` job is **pre-existing on `main` and unrelated to this entry**.

Evidence:
- The failing step is `node scripts/build-site.mjs` (job "Build (locale parity, date derivation, templates)"), which exits 1 with: `no added-date derivable for: https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-ding-sound, …/dsh-wx-push, …/dsh-wx-remote` → `need full git history (fetch-depth: 0) and committed entries — refusing to build`.
- The same `PR check` workflow fails on `main` in runs created 12:04:48, 12:00:53 and 11:57:27 — before this branch existed (PR opened 12:15).
- Those three `wwweljf__dsh-plugins--*` entries are present in `data/plugins/` on `main`, so the failure reproduces without this PR.
- This PR adds exactly one file, `data/plugins/yindf__taskfold.yml`, and the `Submission gate` check passes.

Happy to rebase once `main` is green.

yindf:
The red `check` job is not from this entry — it fails on `main` too, for a pre-existing entry:

```
no added-date derivable for: https://github.com/wwweljf/dsh-plugins/...
```

Our `Submission gate` is green and the submission is the single file `data/plugins/yindf__taskfold.yml`, which touches no other entry. Repo manifest check: `package.json` declares `dsh.bundle.patch` → `./cordis.patch.yml`, the plugin is on npm as `dsh-taskfold@0.31.1`, and a prebuilt tarball is attached to https://github.com/yindf/taskfold/releases/latest. Happy to rebase or adjust the description/category if you'd prefer a different one.

yindf:
Root cause found, in case it helps: the three undated entries come from main commit 5b7be0b9 ("List three submissions that were buried inside stale-fork pull requests", 2026-09-09 03:38Z). Their URLs point at subdirectories of an existing repo (.../tree/master/plugins/dsh-ding-sound), so scripts/build-site.mjs cannot derive an added date from git history and falls back to data/added-dates.json, which is the frozen human-owned baseline.

Adding those three URLs to data/added-dates.json (e.g. 2026-09-09) should turn the build green for every open PR at once — I did not touch it here since it is maintainer-owned. Happy to open that one-line-each PR if you would rather not.

Issue #4731, opened 2026-09-09
build-site: added-date derivation misses entries introduced by merge commits


Mlte0907:
## What happened

`build-site.mjs` exits with code 1:

```
no added-date derivable for: https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-ding-sound, .../dsh-wx-push, .../dsh-wx-remote
need full git history (fetch-depth: 0) and committed entries — refusing to build
```

This blocks **every** PR's `check` job and `main`'s own `build-site.yml` runs since merge commit `5599809c` (2026-09-08, PR #2662).

## Root cause

The three `wwweljf` entry files were introduced **by the merge commit itself**. Both parents lack them:

| ref | `data/plugins/wwweljf__dsh-plugins--plugins-dsh-ding-sound.yml` |
|---|---|
| `57de4187` (first parent) | **404 absent** |
| `34cb8838` (second parent) | **404 absent** |
| `5599809c` (merge) | **present** |

The two date sources in `build-site.mjs` are both `git log` walks:

- **Line 212:** `git log --reverse --date-order --format=%x01%cI -p -- README.md` — reads added README lines
- **Line 241:** `git log --diff-filter=A --format=%cI -- <path>` — reads the file's first commit

`git log` produces no diff for merge commits by default, so **neither source sees the additions**. The script's own error text ("reachable only from a shallow clone or a genuinely uncommitted entry") does not cover this case — it assumes the file was added in a non-merge commit.

Other entries are unaffected because their additions live in the branch's own non-merge commits, which `git log` does diff.

## Suggested fix

Two options:

1. **Pin the three URLs in `data/added-dates.json`** — the quickest fix, zero risk.
2. **Add `--diff-merges=first-parent`** to both `git log` calls — catches all entries going forward.

## Impact

Blocks every contributor whose PR carries a `check` job. The `Submission gate` (entry shape, repo age, `dsh.bundle`) is unaffected — only the site build breaks.

GodCC6:
Independent confirmation, plus **option 1 verified end-to-end** on a clean clone.

Reproduced on an untouched `main` (no PR involved):

```
$ npm ci
$ SKIP_PUBLISH_CHECKS=1 node scripts/build-site.mjs
3431 entries parsed across 2 locales
no added-date derivable for: …/dsh-ding-sound, …/dsh-wx-push, …/dsh-wx-remote
[...]

PR #4743, opened 2026-09-09
build: pin added dates for entries introduced by merge commits

chromoany:
Fixes the build break reported in #4731 (option 1 from that issue: pin the three URLs in `data/added-dates.json`).

`build-site.mjs` derives each entry's added date from two `git log` walks (README lines, then `git log --diff-filter=A -- <entry file>`). Neither sees a file that was introduced **by a merge commit**, because `git log` does not diff merges by default. The three `wwweljf/dsh-plugins` sub-package entries were introduced by merge commit `5599809c` (2026-09-08), so both sources miss them and the build exits 1:

```
no added-date derivable for: …/dsh-ding-sound, …/dsh-wx-push, …/dsh-wx-remote
```

This currently fails on `main` (`build-site.yml` since 2026-09-09T03:13Z) and therefore on every contributor PR's `check` job, so no new entry can be published until it is unblocked.

Change: three keys in the frozen `data/added-dates.json` baseline, dated to that merge commit (`2026-09-08T15:01:39Z`), as verified in the #4731 comment. `data/added-dates.json` is read-only for the build and written by no script, so the pins are inert for every other entry.

Option 2 from the issue (`--diff-merges=first-parent` on both `git log` calls) fixes the class of failure rather than these three instances; this PR is the zero-risk unblock and does not preclude it.

### DIFF
--- a/data/added-dates.json
+++ b/data/added-dates.json
@@ -445,5 +445,8 @@
  "https://github.com/Tabbit-Browser/dsh-tabbit": "2026-08-21",
  "https://github.com/omdsh-dev/stent": "2026-08-13",
  "https://github.com/moguiyu/dsh-tavily/tree/main/packages/dsh-tavily": "2026-08-15",
- "https://github.com/polaris-smart/dsh-devices": "2026-08-19"
+ "https://github.com/polaris-smart/dsh-devices": "2026-08-19",
+ "https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-ding-sound": "2026-09-08T15:01:39.000Z",
+ "https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-wx-push": "2026-09-08T15:01:39.000Z",
+ "https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-wx-remote": "2026-09-08T15:01:39.000Z"
 }


chromoany:
Context from reproducing this independently, which is worth knowing before merging: **these three pins were added twice already and are not on `main` today.**

- `23b8e18` — *fix: pin added-dates for wwweljf/dsh-plugins entries* (2026-09-08T16:35Z)
- `9bee1da` — *Pin added dates for wwweljf entries (#2662)* (2026-09-09T01:23Z)

Both are ancestors of `main` (`compare/9bee1da...main` → `ahead 32, behind 0`), and `data/added-dates.json` at `9bee1da` does contain the three keys — 30,979 bytes there vs 30,665 on `main`, exactly the three lines. The path-history endpoint for `data/added-dates.json` still ends at `ba0889c` (2026-09-07), so the removal is not visible as a change to that file, which fits a merge whose resolution took a pre-pin copy of the file (a stale-fork PR merged after the pin).

So this PR is a **re-pin**, not a first pin. That is also the argument for #4731's option 2 (`--diff-merges=first-parent` on both `git log` calls) as a follow-up: a pin in a data file can be dropped again the same way, while the code fix cannot.

`check` and `Submission gate` are both green on this branch, i.e. `node scripts/build-site.mjs` completes with these three keys present. A dozen-plus PRs opened since 03:13Z are blocked on the same failure.

cc @fkysly — this is the queue unblock, not a plugin submission; happy to switch it to the code change if you would rather take option 2 instead.

chromoany:
One question before this gets merged, because it changes what "unblocks the queue" means here.

The current `check` fails on **280** undated entries, not the three this PR pins — measured on the run that had a patched `build-site.mjs`: [run #7981](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/actions/runs/34474863302). Three of those are the `wwweljf` merge entries; the other 276 are entries whose files were added by ordinary single-parent commits that the yml ledger walk does not resolve. So this PR (and #4709 / #4756, which fix the merge case) will not turn `check` green on their own.

I have the missing data ready: dates for all 279 entries that have an entry file on `main`, each traced to the earliest commit touching its file, with the committer date (`%cI`), and all 279 commits verified one by one (271 `added`, 8 `renamed`, 3 merge, 0 date mismatches). The file and the method are in [this comment on #4731](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/issues/4731#issuecomment-5618736475) — candidate pins, not a submission.

**Want me to add them to this PR?** If yes, say so and I will push a second commit here with all 279 keys (or a separate PR, or nothing at all — your call). If you would rather unblock the queue by fixing the derivation instead, ignore this and I will leave the data where it is; no further comments from me either way.

I am explicitly not opening a competing PR for the same file while this one is open.

chromoany:
**Withdrawing the question above — please do not add those 279 keys to this PR.**

The 279 was measured on my own branch in run #7981 (`fix/diff-merges-for-entry-ledger`), where a missing `--no-patch` made the yml ledger walk return patch output, so it resolved none of the entries the README pass had left. On a pristine clone, `main` reports **3** undated entries; with `--no-patch` (the #4709 form) it reports **0** and the build completes. Both runs plus the per-file `git log` output: [my correction on #4731](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/issues/4731#issuecomment-5620117217) — Mlte0907 found it, I only confirmed it against a full clone.

So this PR stays exactly as it is: the three `wwweljf` pins, a data-only unblock, no second commit from me. Sorry for the detour.

fkysly:

```
```

---

fkysly:
**更正我上一条的一处措辞**：那条说「这条 PR 的内容已经以 `eb665bc93` 进入主干」——对你这条**不准确**。你改的是 `data/added-dates.json`（给那三条钉死日期），而 `eb665bc93` 是改 `scripts/build-site.mjs`。两者解决的是同一个症状，但落地的不是你这份改动。准确的说法是：

- **你的诊断是对的**，而且在那个时间点「钉死日期」确实能让主干立刻变绿——作为应急手段成立；
- **采纳的是另一条路**：从推导逻辑上修（`git log --diff-merges=first-parent --no-patch --diff-filter=A`，且只在默认查询无结果时重试），来自 #4708 / #4709 / #4786 / #4821 / #4871；
- **所以 `added-dates.json` 不需要改了**：它是 2026-08-15 之前的冻结基线加手工 URL 迁移，08-15 之后的条目日期一律由 git 历史自动推导。往里写新条目会让那个「人工账本」逐渐变成半自动的混合物，下次同类 bug 再犯时就更难看出哪些是真的人工决定。

这条 PR 因此关掉——**不是说你判断错了**，是同一个问题已经从根上解决，你这份改动现在是多余的。谢谢你花时间去定位它。

---

**Correcting one line in my previous comment**: it said "this PR's content is on main as `eb665bc93`", which is **not accurate for yours**. You changed `data/added-dates.json` (pinning dates for those three entries); `eb665bc93` changes `scripts/build-site.mjs`. Same symptom, but it is not your diff that landed. Accurately:

- **Your diagnosis was right**, and pinning the dates would genuinely have turned main green at that moment — a valid emergency measure.
- **The route taken was the other one**: fixing the derivation itself (`git log --diff-merges=first-parent --no-patch --diff-filter=A`, retried only when the default lookup finds nothing), from #4708 / #4709 / #4786 / #4821 / #4871.
- **So `added-dates.json` needs no edit**: it is the frozen pre-2026-08-15 baseline plus manual URL migrations, and everything newer derives from git history. Writing new entries into it would slowly turn a human ledger into a half-automatic hybrid, making it harder next time to tell which pins are real human decisions.

Closing this PR on that basis — **not because your call was wrong**, but because the problem is solved at the root and this diff is now redundant. Thanks for taking the time to track it down.

PR #1672, opened 2026-08-18
Add webkubor/dsh-bloom-theme (theme)

webkubor:
Adds [webkubor/dsh-bloom-theme](https://github.com/webkubor/dsh-bloom-theme) under **Themes & Appearance**.

A Morandi palette theme for the DSH Web UI, ported from [Bloom for Typora](https://github.com/webkubor/typora-Bloom-theme). Four OKLCH palettes (mist / cinnabar / petal / ripple), each with light and dark variants, switchable from a top-bar dropdown.

What makes it more than a recolor:

- **Two-track palette** — the original Bloom keeps a darkened "readable" accent for text and buttons, and a separate desaturated Morandi color used only for wide ambient gradients. Porting just the first one turns `petal` from lotus pink into fluorescent magenta, so both tracks are carried over.
- **Texture layer** — ambient gradients, cool hairlines, layered shadows, and Markdown detailing, rather than only remapping DSH's color variables.
- **Accessibility** — all 8 accent/background pairs were computed (OKLCH → sRGB → WCAG) and land at ≥ 4.5:1; the light accents for `petal` and `ripple` were darkened specifically to reach AA.
- **Doesn't cover native controls** — the switcher mounts into the DSH top-bar utilities area next to `Session log` instead of floating above it.

Checklist:

- [x] `package.json` declares `dsh.bundle` (`{ "bundle": { "patch": "./cordis.patch.yml" } }`) with `cordis.patch.yml` at the repo root — verified installable via `dsh plugin add`
- [x] Published to npm as [`@kubor/dsh-bloom-theme`](https://www.npmjs.com/package/@kubor/dsh-bloom-theme)
- [x] One file added under `data/plugins/`, READMEs regenerated with `node scripts/generate-readme.mjs`
- [x] Description matches what the plugin actually does
- [x] MIT licensed, actively maintained

PR #4761, opened 2026-09-10
Update webkubor/dsh-bloom-theme: four variants -> nine

webkubor:
Updating my own entry — it has said "four variants" since #1672 (2026-08-19), and the plugin has grown since.

**What changed**

| | before | now |
|---|---|---|
| variants | four (mist, cinnabar, petal, ripple) | nine (+ sage, stone, lapis, amber, aurora) |
| contrast pairs | eight | 18 |

**How both numbers were checked against the code**

- Nine variants: `VARIANTS` in [`src/palette.ts`](https://github.com/webkubor/dsh-bloom-theme/blob/main/src/palette.ts) lists all nine, each with a full palette definition; `aurora` also has its own motion ramp in `src/tokens.ts`.
- 18 pairs: `npm run check:contrast` on the current tree reports `✓ 18 组配色全部达标` (nine variants × light/dark). It runs on every commit via `npm run check`.

Only `data/plugins/webkubor__dsh-bloom-theme.yml` is touched; both READMEs were regenerated with `scripts/generate-readme.mjs` (no hand edits). Diff is 4 insertions / 4 deletions across the three files.

Branch was synced to upstream `main` before branching, so this isn't a stale-fork PR.


webkubor:
Heads-up on the red `check`: it fails on three entries this PR doesn't touch.

```
no added-date derivable for:
  .../wwweljf/dsh-plugins/tree/master/plugins/dsh-ding-sound
  .../wwweljf/dsh-plugins/tree/master/plugins/dsh-wx-push
  .../wwweljf/dsh-plugins/tree/master/plugins/dsh-wx-remote
```

Those three came in with #2662 on 2026-09-08. The same `build-site.mjs` step is red on `main` (the `PR check` run at 02:22 today), so it looks pre-existing rather than something this change introduced — my diff is 4 insertions / 4 deletions confined to `data/plugins/webkubor__dsh-bloom-theme.yml` plus the two regenerated README lines, and `Submission gate` is green.

Happy to rebase once that's sorted, or to adjust anything about the entry itself.

PR #4796, opened 2026-09-10
build: date entries that a merge commit introduced (#4731)

### LINKED ISSUE #4731: build-site: added-date derivation misses entries introduced by merge commits

chromoany:
Implements option 2 from #4731 as a one-line change, after reproducing both date walks on a synthetic repository so the flag is not merged on faith.

## The change

`scripts/build-site.mjs`, the yml ledger (line 241):

```diff
-        const out = execSync(`git log --diff-filter=A --format=%cI -- ${JSON.stringify(file)}`,
+        const out = execSync(`git log --diff-merges=first-parent --diff-filter=A --format=%cI -- ${JSON.stringify(file)}`,
```

`git log` does not diff merge commits by default, so a path that the **merge itself** introduced has no `A` commit in that walk at all. Those entries keep a missing added-date, `build-site.mjs` reaches its `refusing to build` branch, and exit 1 fails `main` and every open PR — the state since `5599809c`.

## What the flag changes, measured

On a repository built to the same shape (yml + README line added on a side branch, merged with `--no-ff`):

```
$ git log --diff-filter=A --format=%cI -- data/plugins/b__dsh-b.yml
2026-09-01T00:00:00Z                    <- the side branch's own commit (today: found only when it is not merge-introduced)

$ git log --diff-merges=first-parent --diff-filter=A --format=%cI -- data/plugins/b__dsh-b.yml
2026-09-09T12:00:00Z                    <- the merge commit
```

With the flag the merge is the commit that adds the path, so `%cI` is the **merge** date. That is the intended semantics here and not a side effect to paper over: the other ledger already publishes the merge era for these entries — the README line for the three `wwweljf` entries comes from `5b7be0b9` (2026-09-08T15:01:48Z), nine seconds after `5599809c`. The comment added above the call records this, because `%cI` on a merge reads like a bug otherwise.

## The other walk is deliberately untouched

The README walk (line 212) does **not** get the flag. Measured on the same synthetic repo, its result is identical with and without it, and its parser cannot be polluted by merge patches: the loop only reads lines starting with `\x01` or `+` (skipping `+++`), and a merge's `-p` output for the README contains no `+- [name](url) — ` lines — such a line can only come from the side that wrote it, which the same walk already sees through `--reverse --date-order`. Adding the flag there would only widen the merge diffs it scans.

## Relationship to #4743

Independent, either can be taken alone. Worth knowing when sequencing them: with the three keys pinned, `ordered.some((e) => !dates[e.url])` is false on the publishing path, so **neither walk runs at all** — #4743 is the immediate unblock and this is the guard against the same class of failure the next time a submission is merged with `--no-ff`, or the pins are dropped again the way `#2662`'s merge dropped them.

Fixes #4731.

### DIFF
--- a/data/plugins/chromoany__dsh-notify-me.yml
+++ b/data/plugins/chromoany__dsh-notify-me.yml
@@ -1,0 +1,6 @@
+url: https://github.com/chromoany/dsh-notify-me
+name: chromoany/dsh-notify-me
+category: notify
+description:
+  en: Desktop notifications, sounds and a tab-title marker when the agent needs your input or a reply finishes in the background.
+  zh: 模型需要你操作或后台回复完成时，弹出系统通知、提示音并标记标签页标题。
--- a/scripts/build-site.mjs
+++ b/scripts/build-site.mjs
@@ -238,7 +238,14 @@ if (ordered.some((e) => !dates[e.url])) {
       try {
         // Oldest "added" commit for that path. Not `-1`, which git applies
         // before --reverse and would hand back the newest instead.
-        const out = execSync(`git log --diff-filter=A --format=%cI -- ${JSON.stringify(file)}`,
+        // --diff-merges=first-parent: git log does not diff merge commits by
+        // default, so a file the merge itself introduced (a submission merged
+        // with --no-ff, or a batch merge) has no "added" commit at all and the
+        // build exited 1 for that entry, failing main and every open PR. With
+        // the flag the merge is the commit that adds the path, so %cI is the
+        // merge date — which is what the README ledger above already reports
+        // for the same entries (the merge and the README commit are seconds apart).
+        const out = execSync(`git log --diff-merges=first-parent --diff-filter=A --format=%cI -- ${JSON.stringify(file)}`,
           { encoding: 'utf8' }).trim().split('\n').filter(Boolean)
         const iso = out[out.length - 1]
         if (iso) dates[e.url] = new Date(iso).toISOString()


chromoany:
Withdrawing this in favour of #4709 / #4756 — it fixes the wrong thing, and the check run on it says so.

**The run: [#7981](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/actions/runs/34474863302)** — `check` failed in *Build (locale parity, date derivation, templates)*, with **280** undated URLs, not the 3 everyone has been looking at. Ours is one of them.

What the 280 are (verified through the API against `main`):

- **279 have an entry file** under `data/plugins/` matching `slugFor(url)` + `.yml` (the tree has 3434 entry files, no truncation), and 279 of 280 have a README line that matches the parser's regex
- their add commits are **ordinary single-parent commits** — sampled 25, all `status: added`, `parents: 1`, dated 2026-08-27 → 2026-09-08, i.e. before the freeze
- their add commits touch **only** the yml (no README in the same commit, which is expected since sync-readme.yml generates those)

So the yml ledger walk (`git log --diff-filter=A --format=%cI -- <file>`) should resolve all 279 of them and does not. Fixing merge-visibility, which is what this PR and the other two do, leaves that untouched — which is exactly what the run shows: 3 merge-introduced entries out of 280 undated.

Separately, my version of the flag is the weaker one and I would not merge it even for the merge case: `--diff-merges=first-parent` without `--no-patch` also prints the merge's patch. On a synthetic repo the output becomes

```
$ git log --diff-merges=first-parent --diff-filter=A --format=%cI -- new.txt
2026-09-10T20:16:50+08:00
\n diff --git a/new.txt b/new.txt ... +x            <- patch, interleaved with the timestamps
2026-09-10T20:16:50+08:00
\n diff --git a/new.txt b/new.txt ... +x
```

while #4709's form (`--diff-merges=first-parent --no-patch --diff-filter=A --format=%cI`) returns exactly the two timestamps and nothing else. `--no-patch` is not a style preference there, it is what keeps the command's output parseable — the same reason the README walk's parser has to skip `+++`.

#4756 also carries the regression tests (`scripts/added-dates.test.mjs` + a check step) and splits the lookup into `scripts/lib/added-dates.mjs`; both are worth more than this one-liner.

The useful part of my earlier comment on #4731 — that the README walk needs no change, and that the merge-visibility fix is safe — still stands. What it did not have was the run's own count. Closing this rather than leaving a third competing PR in the queue.

PR #4820, opened 2026-09-10
Update LoopX DSH plugin to beta.5

huangruiteng:
## Summary

Update the existing LoopX marketplace entry from beta.4 to beta.5. This release restores GoalBar compatibility on DSH 0.1.5 through its authenticated shared API carrier, while preserving the supported legacy carrier.

## Validation

- Release asset resolves successfully (HTTP 200)
- Downloaded release asset matches the locally packed artifact by SHA-256
- `git diff --check`

Only the existing LoopX entry YAML is changed; generated READMEs are untouched.


huangruiteng:
CI note: the failing site-build step is reproducible on the exact upstream base `5b7be0b95ecaf31e8782d38c41811a863b04432b` before this PR's one-line LoopX update.

Both the base and PR merge ref stop on the same three existing `wwweljf/dsh-plugins` entries because their added dates cannot be derived. This PR does not add, remove, or edit those entries, so I have not bundled an unrelated catalog repair.

Checks scoped to the changed entry pass:

- beta.5 release asset resolves (HTTP 200)
- downloaded asset matches the published package SHA-256
- `node scripts/generate-readme.mjs --check` (3431 entries; both READMEs up to date)
- `git diff --check`

The PR remains a single tarball URL update and is ready for review once the upstream base build is repaired or the check is rerun on a fixed base.

PR #4821, opened 2026-09-10
fix(build): derive added dates through merge commits (unblocks main + PR queue)

drscrewdriver:
## What this fixes

`Build (locale parity, date derivation, templates)` has been failing on **main** since 2026-09-08, on every pull request whose merge ref was built after that, e.g. #4687 ([job log](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/actions/runs/34519562721/job/103013388550)):

```
3433 entries parsed across 2 locales
no added-date derivable for: https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-ding-sound, …/dsh-wx-push, …/dsh-wx-remote
need full git history (fetch-depth: 0) and committed entries — refusing to build
```

Those three entries were introduced **by the merge commit itself**. #2662 was merged as `5599809cf` (2026-09-08T15:01:39Z, `parents=2`), whose resolution renamed the entry files to their slug and wrote the README lines:

| | branch commits | on main |
|---|---|---|
| entry file | `wwweljf__dsh-plugins--dsh-ding-sound.yml` (now 404 on main) | `wwweljf__dsh-plugins--plugins-dsh-ding-sound.yml` |
| README line | none of the 4 branch commits touches a README | added by the merge commit (`README.md` `+3`) |

```
$ gh api "repos/awesome-dsh-plugin/awesome-dsh-plugin/commits?path=data/plugins/wwweljf__dsh-plugins--plugins-dsh-ding-sound.yml"
5599809cf  2026-09-08T15:01:39Z  parents=2   Add wwweljf/dsh-plugins: dsh-ding-sound, dsh-wx-push, dsh-wx-remote (#2662)
```

`git log` does not diff a merge by default, so both passes of the derivation come up empty — README history has no `+` line for those URLs, and the entry file's only commit is the merge. Neither URL is in `data/added-dates.json`, so the build refuses, by design.

## Changes

**`data/added-dates.json`** — pins the three entries to `2026-09-08` (the day they landed, matching the merge commit). This is the manual-migration path the file already exists for, and it makes the published dates independent of how the history is shaped later.

**`scripts/build-site.mjs`** — `--diff-merges=first-parent` on the yml lookup, so a merge counts as the commit that added the path. Verified against a reproduction of this exact history shape:

| | current code | with the flag |
|---|---|---|
| `git log --diff-filter=A -- <entry file>` | `''` | `2026-09-08T15:01:39+08:00` |
[...]

Issue #4807, opened 2026-09-10
pr-check is red for every PR: build-site.mjs cannot derive added-date (wwweljf/dsh-plugins entries)


LAYZR114:
### What happens

`pr-check` fails on every PR, in the last step only:

```
Build (local parity, date derivation, templates)  <-- fails
```

`node scripts/build-site.mjs` prints:

```
3432 entries parsed across 2 locales
no added-date derivable for: https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-ding-sound,
                             https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-wx-push,
                             https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-wx-remote
need full git history (fetch-depth: 0) and committed entries - refusing to build
##[error]Process completed with exit code 1.
```

### Repro

Any open PR, e.g. #4739, #4801, #4802, #4805 — all show `check = failure`. Every other step passes (`Stale-fork guard`, `Every entry file ends in .yml`, `Entry files live in data/plugins/`, `READMEs match data/plugins`, `awesome-lint`), and `Submission gate` passes for most of them ("Entries look good").

Since the three entries above live in `main`, the build refuses for every PR — no submission can go green until this is fixed.

### Likely cause

Those three entries are monorepo-subpackage entries (`name: wwweljf/dsh-plugins#dsh-ding-sound`, `url: .../tree/master/plugins/dsh-ding-sound`). If the added-date comes from something like `git log --diff-filter=A -- data/plugins/<file>.yml`, it returns nothing for files that arrived through a bulk import or a rename.

### Suggested fixes (any one unblocks all PRs)

1. Use `git log --follow --diff-filter=A` so renames are followed.
2. Fall back to the file's first appearance in any ref (`--all`), or to the commit that last changed the file.
3. Degrade to a warning for entries without a derivable date instead of aborting the whole build.

Happy to test a fix with a PR if useful.

cjm-m:
Independent confirmation from a blocked submission — my PR #4695 shows the exact same split verdict:

- `Submission gate`: **success** ("All 1 submitted entry passes: dsh.bundle declared, repo old enough, enough commits")
- `check` (pr-check): **failure**, identical message — `no added-date derivable for: ...wwweljf/dsh-plugins/tree/master/plugins/dsh-ding-sound, dsh-wx-push, dsh-wx-remote`

[...]

PR #4786, opened 2026-09-10
Add SodaZheng/dsh-totp to security plugins

SodaZheng:
Adds [SodaZheng/dsh-totp](https://github.com/SodaZheng/dsh-totp) to **Security & Permissions**.

The plugin provides TOTP-only access control for personal DeepSeek Harness Web instances, with QR enrollment in DSH Settings, single-use recovery codes, live protection controls, and lock-all revocation of page access. Protection is off on a fresh installation and turns on after enrollment. This is not password-plus-TOTP two-factor authentication or multi-user authorization.

Compared with existing authentication gateways, this submission focuses on the combination of TOTP-only sign-in, in-app enrollment/replacement that keeps the enrolling page usable, and live protection/lock-all controls.

Validation: `npm run verify` passes 45 tests and package/bundle/client-registration checks. The packaged plugin was installed and started in an isolated DSH `0.1.2-rc.1` profile on macOS; HTTP bootstrap, frontend scripts, and trusted-host handling were checked. Published on npm as `dsh-totp@0.0.3`.

- [x] Added only `data/plugins/SodaZheng__dsh-totp.yml`; no generated README or unrelated entry changes.
- [x] The plugin's `package.json` declares `dsh.bundle`.
- [ ] Repository is at least 1 day old: created at **2026-09-09 09:36:51 UTC**, so this becomes true at **2026-09-10 09:36:51 UTC**. The repository-age check can clear on the project's automatic recheck; no age override is requested.
- [x] Category is `security`.
- [x] Description states implemented functionality without superlatives.
- [x] Repository has the `dsh-plugin` topic.

Official `@deepseek-ai/*` packages are peer dependencies. Source and review evidence: [plugin conventions](https://github.com/SodaZheng/dsh-totp/blob/main/docs/plugin-conventions.md), [submission notes](https://github.com/SodaZheng/dsh-totp/blob/main/docs/submission/README.md).

PR #4831, opened 2026-09-11
Add yunxiyang/dsh-loop-continue

yunxiyang:
Adds `yunxiyang/dsh-loop-continue`.

## What it does

The agent loop closes a turn as `completed` the moment a step emits text and no tool call. A model that writes "now I will re-apply the change" and stops therefore looks *finished* to the loop, even though the work is half done.

`dsh-loop-continue` listens on `agent/turn-stopping`, replays the stopping turn's own session log, and asks a judge model one strict `true`/`false` question — but only when the turn shape matches that failure mode:

1. the turn ended on a text-only step (no tool call), **and**
2. it called at least one tool earlier.

So an ordinary finished answer, or a one-shot reply that never used a tool, never spends a judge call. On `true` the guard steers the same turn to run one more step; on `false` (or an unparseable answer) the turn closes as usual.

## Bounds

Steering is not free, so it is capped two ways:

- a hard per-turn budget (`maxContinuations`, default 10), and
- an **ignored-steer check**: the guard records the session-log position at each steer and, the next time the turn stops, verifies that some step in between actually called a tool. A steer the model ignored ends the turn instead of spending the rest of the budget on the same refusal.

The judge follows the conversation's own provider/model and is advisory — a judge failure never wedges a turn.

## Manifest

`package.json` declares `dsh.bundle` with `./cordis.patch.yml`, which ships a plain `id` + `name` insert (no `config`, no expressions), and the repo has the `dsh-plugin` topic. Published to npm as `dsh-loop-continue` (0.1.1).

## Entry

Adds one file, `data/plugins/yunxiyang__dsh-loop-continue.yml`. `scripts/check-submission.mjs` passes for it locally.

## Relationship to nearby entries

Three entries already act on `agent/turn-stopping`; none of them covers this failure mode.

- `Milbaxter/dsh-critique-loop` intercepts **every** completed turn to force a critique round. This plugin is narrower: it fires only when the turn's shape shows the model announced an action and never took it, so a finished turn is left alone.
- `pavangupta352/stalegreen` (#dsh-plugin-stalegreen) steers when a test/lint/build *claim* has a stale or failed receipt — a freshness check on verification claims. This one never looks at claim freshness; it looks at whether a described action was ever performed.
- `CAI-MH/dsh-quality-review` audits the *quality* of a reply (accuracy, completeness, consistency).

The distinguishing case here is a control-flow defect, not a review: the loop closes the turn as `completed` because the step carried text and no tool call, and the promise inside that text is silently dropped.

### DIFF
--- a/data/added-dates.json
+++ b/data/added-dates.json
@@ -445,5 +445,8 @@
  "https://github.com/Tabbit-Browser/dsh-tabbit": "2026-08-21",
  "https://github.com/omdsh-dev/stent": "2026-08-13",
  "https://github.com/moguiyu/dsh-tavily/tree/main/packages/dsh-tavily": "2026-08-15",
- "https://github.com/polaris-smart/dsh-devices": "2026-08-19"
+ "https://github.com/polaris-smart/dsh-devices": "2026-08-19",
+ "https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-ding-sound": "2026-09-08",
+ "https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-wx-push": "2026-09-08",
+ "https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-wx-remote": "2026-09-08"
 }
--- a/data/plugins/yunxiyang__dsh-loop-continue.yml
+++ b/data/plugins/yunxiyang__dsh-loop-continue.yml
@@ -1,0 +1,6 @@
+url: https://github.com/yunxiyang/dsh-loop-continue
+name: yunxiyang/dsh-loop-continue
+category: agi
+description:
+  en: 'Resumes an agent turn that ended after narrating its next action without calling a tool: the guard replays the turn log on agent/turn-stopping, asks a judge model one true/false question, and steers the same turn to run one more step.'
+  zh: '续跑「叙述了下一步动作却没调用工具」就结束的 agent 轮次：插件在 agent/turn-stopping 回放该轮日志，让裁判模型回答一次 true/false，命中的话把同一轮再推进一步。'


yunxiyang:
Heads-up on the red `check`: the failing step is `Build (locale parity, date derivation, templates)` → `node scripts/build-site.mjs`, which aborts with

```
```

This is not from this submission — those three are `wwweljf/dsh-plugins` subpackage entries. I reproduced it on a clean `main` checkout (3431 entries, none of my changes) and got the identical failure, which matches #4807 / #4731.

The `Submission gate` job passed, and `scripts/check-submission.mjs --base HEAD~1` passes locally for this entry:

```
checking 1 entry
  ok  https://github.com/yunxiyang/dsh-loop-continue
all checked entries pass
```

The PR is the single added file `data/plugins/yunxiyang__dsh-loop-continue.yml`, +6/-0.

drscrewdriver:
是否先试下加修复？ https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/pull/4821 try it
example pr#4687

yunxiyang:
Done — thanks for the pointer.

I took the `#4687` route rather than the `#4821` script fix, since it is the smaller change and the build only ever named those three URLs:

```
data/added-dates.json | +4 -1
```

```diff
- "https://github.com/polaris-smart/dsh-devices": "2026-08-19"
+ "https://github.com/polaris-smart/dsh-devices": "2026-08-19",
+ "https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-ding-sound": "2026-09-08",
+ "https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-wx-push": "2026-09-08",
+ "https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-wx-remote": "2026-09-08"
 ```

`check` is green now (and so is `Submission gate`), `mergeStateStatus: CLEAN`. The entry itself is unchanged: `data/plugins/yunxiyang__dsh-loop-continue.yml`, +6/-0.

Worth noting the underlying reason those three are unreachable by derivation, in case it matters for `#4821`: #2662 was merged as `5599809cf` (`parents=2`), and its resolution both renamed the entry files to their slug and wrote the README lines. So the entry file's only history is the merge commit, and the README lines exist in no non-merge commit — `git log` shows no diff for a merge by default, which is exactly what `--diff-merges=first-parent` addresses.

I left `scripts/build-site.mjs` alone here to keep this PR to one added file plus the ledger; happy to fold in the `#4821` change instead if you would rather land them together.

fkysly:
已合入 — 你的条目在列表里了。**这个 PR 里为绕过构建失败而加的那半没有一起合**，原因是那个 bug 已经从根上修掉了。

**先说明你当时看到的红是怎么回事，那不是你的问题。** 9-08 我合 [#2662](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/pull/2662) 时，把它的三个条目文件改名以匹配各自的 url，并在同一个合并提交里重新生成了 README 行。`build-site.mjs` 推导「收录日期」时用 `git log --diff-filter=A`，而**合并提交没有单一 diff，git 默认不会为它计算**——于是那三条在文件账本和 README 账本里同时失明，构建按设计拒绝运行（盖一个"现在"会让日期每次构建都跳变）。

后果是主干从那一刻红到今天：站点冻结在 3408 条，而**每个开着的 PR 都在 `check` 这一步继承了这个失败**，130 个与此毫无关系的投稿跟着变红。你在 PR 里顺手补日期或改构建脚本，是在替我擦这个屎，谢谢。

**修法来自投稿人，不是我。** [#4708](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/pull/4708)、[#4709](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/pull/4709)、[#4786](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/pull/4786)、[#4821](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/pull/4821)、[#4871](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/pull/4871) 五份各自独立诊断出了同一个原因，三份收敛到 `git log --diff-merges=first-parent --no-patch --diff-filter=A`，并且**只在默认查询一无所获时才重试**。我自己先推的 `--first-parent` 版本把 170 个条目的日期错误地挪到了合并日，已回滚换成他们的（`eb665bc93`）。

**所以 `data/added-dates.json` 不需要补**：它是 2026-08-15 之前的冻结基线加手工 URL 迁移，之后的日期全部由 git 历史自动推导，你的条目日期现在是自动算出来的真实合入日。`scripts/build-site.mjs` 的修复也已在主干。

---

Merged — your entry is listed. **The half this PR carried to work around the broken build was not merged**, because the bug is fixed at the root now.

**First, the red you saw was not yours.** Merging [#2662](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/pull/2662) on Sep 8 renamed its three entry files to match their urls and regenerated their README lines inside the same merge commit. `build-site.mjs` derives an entry's added date with `git log --diff-filter=A`, and **a merge has no single diff, so git computes none for it by default** — both ledgers went blind for those three, and the build refused to run, correctly, since stamping "now" would make dates flap between builds.

main was red from that moment: the site froze at 3,408 entries and **every open pull request inherited the failure on the `check` step** — 130 unrelated submissions turned red. Patching the date file or the build script inside your own PR was you cleaning up after me. Thank you for it.

**The fix is the contributors', not mine.** #4708, #4709, #4786, #4821 and #4871 each diagnosed the same cause independently, three converged on `git log --diff-merges=first-parent --no-patch --diff-filter=A` retried **only when the default lookup finds nothing**, and that is what is on main. My own first attempt (`--first-parent`) wrongly moved 170 entries' dates to their merge date and was reverted in favour of theirs (`eb665bc93`).

**So `data/added-dates.json` needs no backfill**: it is the frozen pre-2026-08-15 baseline plus manual URL migrations, and everything newer derives from git history — your entry's date is now computed automatically from its real merge commit.

yunxiyang:
Merged — thanks for the write-up, and for tracking down what actually happened.

Good to know the backfill was unnecessary: I had read the three URLs as a ledger gap rather than as a symptom. `--diff-merges=first-parent` retried only on an empty default lookup is the right shape — it leaves every date that already derived correctly alone, which is exactly where the first `--first-parent` attempt went wrong.

`dsh-loop-continue` 0.2.0 is out on npm with the settings card; nothing in it touches this repo.

PR #4845, opened 2026-09-11
Add WYR-233/dsh-multi-tts (voice)

WYR-233:
**Plugin**: https://github.com/WYR-233/dsh-multi-tts

Adds one entry file `data/plugins/WYR-233__dsh-multi-tts.yml` (voice).

- `dsh.bundle.patch` is declared in package.json and the repo ships `cordis.patch.yml`.
- Install: `dsh plugin --profile web add github:WYR-233/dsh-multi-tts`; a prebuilt tarball is attached to the latest GitHub Release (asset name carries no version).
- `@deepseek-ai/*` packages are declared as peerDependencies with stable ranges.
- `dsh-plugin` topic is set on the repository.

What it does: adds a read-aloud button to every assistant reply plus an auto-read toggle in the composer, backed by a settings page where the user picks the TTS provider (MiniMax, or any OpenAI-compatible `/audio/speech` endpoint), voice, emotion, speed and model. Text is cleaned of code blocks, links, paths and Markdown markup before synthesis; API keys resolve from plugin config, environment variables or `$DSH_HOME/.env`. Config writes are atomic and keep a one-shot backup, and a corrupt config self-heals instead of being silently wiped.

_(This PR previously also carried a second plugin; that submission now lives in its own PR #4879.)_

### DIFF
--- a/data/plugins/WYR-233__dsh-memory-lite.yml
+++ b/data/plugins/WYR-233__dsh-memory-lite.yml
@@ -1,0 +1,7 @@
+url: https://github.com/WYR-233/dsh-memory-lite
+name: WYR-233/dsh-memory-lite
+category: memory
+description:
+  en: 'Read-only enhancer for a file-based long-term memory library (markdown cards in git): a memory_find search tool, a small budgeted memory digest injected into the system prompt, and a settings page with library overview, health-check status and toggles.'
+  zh: '面向「纯文件（markdown 卡片 + git）长期记忆库」的只读增强插件：memory_find 检索工具、一小段受预算约束的记忆库摘要注入，以及带概览/健康/开关的设置页。'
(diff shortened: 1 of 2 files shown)


WYR-233:
Heads-up on the failing `check` job (it is not about this entry):

```
no added-date derivable for: https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-ding-sound,
  .../dsh-wx-push, .../dsh-wx-remote
```

Those three URLs are other people's entries, and the build refuses to proceed for the whole site, so every submission `check` run dies the same way right now (10 of the last 12 PRs show the same red `check`). The entry in this PR parses fine — it is plain `url/name/category/description/tarball` with quoted descriptions.

Happy to rebase/re-run once the upstream data is fixed, or if a maintainer prefers I can open a minimal PR that adds the missing added-date source for those three URLs. Just say the word.

PR #4853, opened 2026-09-11
Restore wwweljf added-date pins lost to a stale-base merge

tay3223:
9bee1da31 pinned the added dates for the three #2662 entries (merge-commit introductions the git-history derivation can't see). That pin is **gone from current main**: a later merge whose branch predated 9bee1da31 rewrote `data/added-dates.json` from a stale base, dropping the three keys. The reversal hides inside a merge commit, so the path-filtered history shows nothing after 09-07.

Effect: every PR based on current main fails the Build step with `no added-date derivable for: …/dsh-ding-sound, …/dsh-wx-push, …/dsh-wx-remote` — #4682, #4683 and the rest of the open pile.

This PR restores the exact file content from 9bee1da31 (verified: the only delta between that commit and main on this path is the three lost keys). `SKIP_PUBLISH_CHECKS=1 node scripts/build-site.mjs` passes locally on this branch and fails on main.

Possible prevention (not included here): the added-dates.json write path could assert it never drops existing keys, or the derivation walk could expand merges (`-m`) — either would make this class of reversal loud instead of silent.

### DIFF
--- a/data/added-dates.json
+++ b/data/added-dates.json
@@ -445,5 +445,8 @@
  "https://github.com/Tabbit-Browser/dsh-tabbit": "2026-08-21",
  "https://github.com/omdsh-dev/stent": "2026-08-13",
  "https://github.com/moguiyu/dsh-tavily/tree/main/packages/dsh-tavily": "2026-08-15",
- "https://github.com/polaris-smart/dsh-devices": "2026-08-19"
+ "https://github.com/polaris-smart/dsh-devices": "2026-08-19",
+ "https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-ding-sound": "2026-09-08T15:01:39.000Z",
+ "https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-wx-push": "2026-09-08T15:01:39.000Z",
+ "https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-wx-remote": "2026-09-08T15:01:39.000Z"
 }


fkysly:

```
```

---

fkysly:
**更正我上一条的一处措辞**：那条说「这条 PR 的内容已经以 `eb665bc93` 进入主干」——对你这条**不准确**。你改的是 `data/added-dates.json`（给那三条钉死日期），而 `eb665bc93` 是改 `scripts/build-site.mjs`。两者解决的是同一个症状，但落地的不是你这份改动。准确的说法是：

- **你的诊断是对的**，而且在那个时间点「钉死日期」确实能让主干立刻变绿——作为应急手段成立；
- **采纳的是另一条路**：从推导逻辑上修（`git log --diff-merges=first-parent --no-patch --diff-filter=A`，且只在默认查询无结果时重试），来自 #4708 / #4709 / #4786 / #4821 / #4871；
- **所以 `added-dates.json` 不需要改了**：它是 2026-08-15 之前的冻结基线加手工 URL 迁移，08-15 之后的条目日期一律由 git 历史自动推导。往里写新条目会让那个「人工账本」逐渐变成半自动的混合物，下次同类 bug 再犯时就更难看出哪些是真的人工决定。

这条 PR 因此关掉——**不是说你判断错了**，是同一个问题已经从根上解决，你这份改动现在是多余的。谢谢你花时间去定位它。

---

**Correcting one line in my previous comment**: it said "this PR's content is on main as `eb665bc93`", which is **not accurate for yours**. You changed `data/added-dates.json` (pinning dates for those three entries); `eb665bc93` changes `scripts/build-site.mjs`. Same symptom, but it is not your diff that landed. Accurately:

- **Your diagnosis was right**, and pinning the dates would genuinely have turned main green at that moment — a valid emergency measure.
- **The route taken was the other one**: fixing the derivation itself (`git log --diff-merges=first-parent --no-patch --diff-filter=A`, retried only when the default lookup finds nothing), from #4708 / #4709 / #4786 / #4821 / #4871.
- **So `added-dates.json` needs no edit**: it is the frozen pre-2026-08-15 baseline plus manual URL migrations, and everything newer derives from git history. Writing new entries into it would slowly turn a human ledger into a half-automatic hybrid, making it harder next time to tell which pins are real human decisions.

Closing this PR on that basis — **not because your call was wrong**, but because the problem is solved at the root and this diff is now redundant. Thanks for taking the time to track it down.

Issue #4847, opened 2026-09-11
`pr-check` → **Build (locale parity, date derivation, templates)** fails for **every** PR


LittleBlackTong:
### What kind of problem? / 问题类型

awesome-dsh-plugin.com is broken / 网站有问题

### Which entry? / 哪个条目

https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-ding-sound

### What's wrong? / 具体问题

`pr-check` → **Build (locale parity, date derivation, templates)** fails for **every** PR on current `main` (`5b7be0b`):

    no added-date derivable for:
      https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-ding-sound
      https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-wx-push
      https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-wx-remote
    need full git history (fetch-depth: 0) and committed entries — refusing to build

Reproduce locally: `npm ci && SKIP_PUBLISH_CHECKS=1 node scripts/build-site.mjs`

Both derivation passes miss these three entries:

1. README history (`git log --reverse -p -- README.md`) — `git log -p` does not emit diffs for merge commits by default. Their README lines were introduced by a merge commit (`5599809cf`, "Add wwweljf/dsh-plugins … (#2662)").
2. yml fallback (`git log --diff-filter=A -- <file>`) — the same commit renamed the three files (`wwweljf__dsh-plugins--dsh-*.yml` → `wwweljf__dsh-plugins--plugins-dsh-*.yml`); `--diff-filter=A` does not follow renames, so there is no `A` record for the current paths.

They are not in `data/added-dates.json` either, so nothing short-circuits the derivation.

Possible fixes (either):
- data: add the three URLs to `data/added-dates.json`;
- code: make the derivation merge/rename-aware (e.g. `--first-parent -m` for the README pass; `--follow` / `--diff-filter=ACR` for the yml fallback).

Hit while submitting #4746 and #4721 (both otherwise clean, `Submission gate` green). Blocks the whole PR queue, not just those two.

drscrewdriver:
+1

drscrewdriver:
需要类似修复吧https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/pull/4821

fkysly:
[...]

PR #4855, opened 2026-09-11
Add dsh-opencode-session

nobu121:
Adds one entry under **Models & Providers**.

- Repo: https://github.com/nobu121/dsh-opencode-session
- What it does: attaches a stable, per-conversation `x-opencode-session` header to model requests routed to OpenCode / OpenCode Go provider routes, which fixes `400 MissingSessionID` and keeps the session affinity / prompt-cache routing that a stable header value provides. The default value is the DSH session id of the call; a `uuid` mode is also available. Other providers, requests without a session id, and requests that already carry the header are left untouched.
- Manifest: `dsh.bundle.patch` -> `cordis.patch.yml` at the repo root. Zero runtime dependencies; `npm test` (9 behavior tests against an echo server) passes.
- Published on npm as `dsh-opencode-session`; 0.1.1 adds the `repository` field pointing back at the repo.

Note: the repository was created today, so the 1-day repo-age check will fail on this run. I will re-trigger CI once the repo is past 24 hours.


nobu121:
CI is red on this PR, but not because of this entry.

The failing step is the known `scripts/build-site.mjs` added-date derivation problem that is currently failing **every** PR — upstream issues #4847 and #4807, both opened before this PR. The output here is byte-identical to the failure on `Add vclike/dsh-github-companion` (run 34575512723):

```
```

Every other gate in this run passed: Stale-fork guard, "Every entry file ends in .yml", "Entry files live in data/plugins/", "READMEs match data/plugins", and awesome-lint.

Nothing to change on my side. I will re-run the check once the build is fixed upstream — and, since the repository was created today, after it is past the 1-day age gate (2026-09-12 ~07:36 UTC).

nobu121:
CI is green now (run 34802403210, 7m21s): stale-fork guard, entry filename/location checks, README sync, awesome-lint, added-date regression tests, and the site build all pass.

Root cause of the earlier red run: the branch was based on a pre-fix `main`, so the `READMEs match data/plugins` step (which diffs against the merge-base) saw three days of upstream README churn as "PR changes". Rebased onto current `main` — the PR now carries exactly one file, +1/-0.

PR #4792, opened 2026-09-10
Fix added-date derivation for the three wwweljf/dsh-plugins entries (unblocks the PR check build)

icanfinish11:
## What

`data/added-dates.json` is missing the three `wwweljf/dsh-plugins` entries that were added on 2026-09-08 (commit `5599809c`, "Add wwweljf/dsh-plugins: dsh-ding-sound, dsh-wx-push, dsh-wx-remote (#2662)"). This pins their dates in the frozen baseline: **one key given a comma plus three added lines, nothing else.**

## Why

`scripts/build-site.mjs` refuses to build when an entry has no derivable added date:

```
no added-date derivable for: https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-ding-sound,
  https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-wx-push,
  https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-wx-remote
need full git history (fetch-depth: 0) and committed entries — refusing to build
```

The derivation resolves an entry's date from the commit that first added it (`git log --diff-filter=A -- <entry file>`, with a README-history walk in front of it). For these three it yields nothing — their entry files were renamed to match their URLs in that same commit — so the build exits 1.

That makes `PR check` red on every open pull request right now, on exactly these three URLs, while the `Submission gate` results differ normally per PR. Sampled at the time of writing: #4777–#4788 and #4790 all fail on this same step.

`data/added-dates.json` is documented in `build-site.mjs` as the human-owned baseline for precisely this situation:

> `data/added-dates.json` is a frozen, human-owned baseline: it pins the dates published before 2026-08-15 and carries manual migrations (an entry repointed to a new URL keeps its original date).

So pinning the three dates here is the smallest change that lets the build run again; no workflow change is needed (`pr-check.yml` already uses `fetch-depth: 0`).

## The dates

All three entries arrived in one commit — `5599809c`, committer date `2026-09-08T15:01:39Z` — so each is pinned to `2026-09-08`. Happy to switch to the timestamped form (`2026-09-08T15:01:39.000Z`, which some rows use) if you prefer it; just say so and I will push to this branch.

## Scope

- One data file changed: `data/added-dates.json` (+4/-1).
- No entry file is touched, the READMEs are untouched, no other data is touched, and nothing in the file was reordered or reformatted (LF endings preserved; the result is byte-identical to `main` apart from those lines).

## Context

[...]

PR #4857, opened 2026-09-11
Add zhqowo/dsh-whale-tray (ui)

zhqowo:
Adds **zhqowo/dsh-whale-tray** — a tray / menu-bar launcher for DeepSeek Harness (Windows system tray + macOS menu bar).

One file, `data/plugins/zhqowo__dsh-whale-tray.yml`, category `ui` (same as the other desktop-shell / tray entries).

### Installability checked before submitting

- `package.json` declares `dsh.bundle.patch` → `./cordis.patch.yml`
- `cordis.patch.yml` is present at the repo root
- `pnpm add github:zhqowo/dsh-whale-tray` installs cleanly; the installed package carries both the manifest and the patch file

### A note on the patch

The patch is intentionally a **no-op** (`[]`). What actually does the work is a compiled desktop binary (WinForms `.exe` / Swift `.app`) which cannot be distributed over npm — the binaries ship from the repo's GitHub Releases. The plugin entry is therefore a discovery/registration entry point, not the app itself. Flagging it explicitly in case that is outside what this list wants to carry; happy to adjust or withdraw if so.

The repo also has a macOS sibling launcher for OpenClaw (zhqowo/openclaw-menubar) — deliberately **not** submitted here, as it is not a DSH plugin.


zhqowo:
Heads-up: the `PR check` failure on this PR looks like the known pre-existing `wwweljf` added-date problem, not this submission. Flagging it so nobody has to re-diagnose it.

Evidence it is unrelated to this entry:

- the build error lists exactly three URLs, all `wwweljf/dsh-plugins/...` — the entry added here is not among them
- `zhqowo` does not appear anywhere in the job log
- other PRs opened in the same window fail too (07:24 / 07:32 / 07:40 UTC all `failure`); the only green run nearby is `restore-wwweljf-added-dates`

Reproduced locally against a full clone, using the exact command from `scripts/build-site.mjs:241`:

```
$ git log --diff-filter=A --format=%cI -- data/plugins/zhqowo__dsh-whale-tray.yml
2026-09-11T16:13:34+08:00                      # derives fine

$ git log --diff-filter=A --format=%cI -- data/plugins/wwweljf__dsh-plugins--plugins-dsh-ding-sound.yml
                                                # empty — the three reported entries
```

Root cause, in case it is useful beyond the date restore in #4792 / #4853: those three files were introduced by a **merge commit** (`5599809cf`, 2 parents), and `git log` does not diff merge commits by default, so `--diff-filter=A` has nothing to match against. Adding `-m --first-parent` resolves it:

```
$ git log -m --first-parent --diff-filter=A --format=%cI -- data/plugins/wwweljf__dsh-plugins--plugins-dsh-ding-sound.yml
2026-09-08T23:01:39+08:00
```

That would also cover any future entry that first lands via a merge, which pinning dates in `data/added-dates.json` would not.

Nothing needed from me on this PR — happy to rebase once either fix lands.

PR #4879, opened 2026-09-11
Add WYR-233/dsh-memory-lite (memory)

WYR-233:
**Plugin**: https://github.com/WYR-233/dsh-memory-lite

Adds one entry file `data/plugins/WYR-233__dsh-memory-lite.yml` (memory).

- `dsh.bundle.patch` is declared in package.json and the repo ships `cordis.patch.yml`.
- Install: `dsh plugin --profile web add github:WYR-233/dsh-memory-lite`; a prebuilt tarball is attached to the latest GitHub Release (asset name carries no version).
- `@deepseek-ai/cordis` is a peerDependency with a stable range; the plugin has zero runtime dependencies.
- `dsh-plugin` topic is set on the repository.

What it does: points at any markdown-card long-term memory library and gives the agent a `memory_find` tool plus a small, budgeted digest section in the system prompt. A settings page shows card count, body size, topics, the last health-check result, and toggles for the digest, the tool, the budget and the search limit. It never writes to the library and never rewrites conversation history; the only state is one small JSON config written atomically.

### DIFF
--- a/data/plugins/WYR-233__dsh-memory-lite.yml
+++ b/data/plugins/WYR-233__dsh-memory-lite.yml
@@ -1,0 +1,7 @@
+url: https://github.com/WYR-233/dsh-memory-lite
+name: WYR-233/dsh-memory-lite
+category: memory
+description:
+  en: 'Read-only enhancer for a file-based long-term memory library (markdown cards in git): a memory_find search tool, a small budgeted memory digest injected into the system prompt, and a settings page with library overview, health-check status and toggles.'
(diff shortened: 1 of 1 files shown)


WYR-233:
Heads-up on the failing `check` job (it is not about this entry):

```
no added-date derivable for: https://github.com/wwweljf/dsh-plugins/tree/master/plugins/dsh-ding-sound,
  .../dsh-wx-push, .../dsh-wx-remote
```

Those three URLs are other people's entries, and the build refuses to proceed for the whole site, so every submission `check` run dies the same way right now (10 of the last 12 PRs show the same red `check`). The entry in this PR parses fine — it is plain `url/name/category/description/tarball` with quoted descriptions.

Happy to rebase/re-run once the upstream data is fixed, or if a maintainer prefers I can open a minimal PR that adds the missing added-date source for those three URLs. Just say the word.

PR #4881, opened 2026-09-11
Add dsh-mobile-direct (category: remote)

hajimimaodie8:
Adds one entry file, `data/plugins/hajimimaodie8__dsh-mobile-direct.yml`, per contributing.md.

`dsh-mobile-direct` puts a QR in the top-right of the harness web UI so a phone can get in without typing anything:

- the QR carries a `dsh-relay` pairing payload (`kind: dsh-relay-pair`, `v: 1`) whose `url` is the **LAN address** the phone can actually reach. The relay builds that URL from the `Host` you opened its page with, so the QR it renders on the machine running the harness says `127.0.0.1`, which no phone can use;
- the DSH Mobile app scans it and pairs via `POST /relay/pair`, receiving a device token (30 days by default);
- a second QR opens a one-tap browser session for phones without the app.

Repo: https://github.com/hajimimaodie8/dsh-mobile-direct
Requires `dsh-relay` >= 0.2.0. No build step; the package ships plain ESM.

### DIFF
--- a/data/plugins/hajimimaodie8__dsh-mobile-direct.yml
+++ b/data/plugins/hajimimaodie8__dsh-mobile-direct.yml
@@ -1,0 +1,6 @@
+url: https://github.com/hajimimaodie8/dsh-mobile-direct
+name: hajimimaodie8/dsh-mobile-direct
+category: remote
+description:
+  en: 'Scan-once mobile entry 鈥?a QR in the top-right of the harness web UI that pairs the DSH Mobile app (device token, 30 days) or opens a one-tap browser session, always carrying a LAN address the phone can actually reach.'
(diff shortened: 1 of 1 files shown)


hajimimaodie8:
Withdrawing this submission: the plugin is not safe yet.

On the maintainer's own machine it twice drove the harness to a fatal heap exhaustion
(`FATAL ERROR: Ineffective mark-compacts near heap limit ... JavaScript heap out of memory`,
then `exit code 134`), which put the desktop shell into its safe mode. The first hypothesis
(state flushed to `$DSH_HOME` on every authenticated request, feeding the desktop shell's
config watcher) was fixed, but the second crash reproduced anyway, so the real cause is still
unknown. It has been uninstalled locally and the PR is closed until it can be reproduced and
fixed in an isolated harness.

Sorry for the noise 鈥?I will reopen with evidence once it is actually safe.

PR #4908, opened 2026-09-12
Add dsh-creative-studio - portable image + video skill pack

Ylight-cpu:
Adds one entry: `data/plugins/Ylight-cpu__dsh-creative-studio.yml` (category `skill`).

## What the plugin does

`dsh-creative-studio` is a skill pack that gives the agent real image and video production capability through a bundled Python toolkit (`toolkit/cs.py`, 18 subcommands) plus two skills:

- **image-studio** — background removal with a quality policy (`fast` u2net / `auto` isnet-or-BiRefNet by image size / `best` BiRefNet), local inpainting retouch (`inpaint` with mask or box/circle regions), headline and logo compositing, multi-layer `compose`, platform-size batch exports (amazon-main, shopify, faire, 9:16 …), contact sheets, palette extraction and an image QA report (`inspect` reports subject coverage and edge transition width).
- **video-studio** — footage inventory (`video-probe`), frame extraction, accurate or stream-copy cutting, transition joins with timebase/fps/pixel-format normalisation, on-screen text, subtitle burn-in, music mixing with fades, sharpest-frame cover selection, and platform-spec exports.

The Python runtime is not bundled in the package; `scripts/bootstrap.ps1` provisions it per machine (Pillow, OpenCV, rembg, ONNX Runtime, imageio-ffmpeg) and `scripts/install-as-plugin.ps1` registers the plugin.

## Checks

- `package.json` declares `dsh.bundle.patch` → `./cordis.patch.yml`; the patch inserts the plugin row.
- Installs from the published source: `dsh plugin --profile web add github:Ylight-cpu/dsh-creative-studio` was run against a throwaway profile and completed, with `dsh-creative-studio` reconciled into `dsh.profile.bundles`.
- Prebuilt tarball attached to the v0.1.0 release and referenced from the entry with a version-free asset name; `https://github.com/Ylight-cpu/dsh-creative-studio/releases/latest/download/dsh-creative-studio.tgz` was fetched and verified (62.3 KB).
- Repository topic `dsh-plugin` added.
- Official `@deepseek-ai/*` packages are declared as `peerDependencies` with an explicit prerelease branch (`>=0.1.1-rc.2 <0.1.2 || >=0.1.2-rc.1 <0.2.0-0`), verified with semver against `0.1.2-rc.1`.

The repository was created today, so the age check may report "repository is 0.x days old (needs 1)" on the first run — flagging it here so it is not read as something to fix; the entry itself is ready.

### DIFF
--- a/data/plugins/Ylight-cpu__dsh-creative-studio.yml
+++ b/data/plugins/Ylight-cpu__dsh-creative-studio.yml
@@ -1,0 +1,23 @@
+# Submission file for the awesome-dsh-plugin list.
+#
+# Destination in the target repository:
+#   data/plugins/Ylight-cpu__dsh-creative-studio.yml
+#
+# Rules this file follows (from contributing.md):
+#   * one YAML file per plugin; the READMEs are generated — never edit them by hand
+#   * description.en is required and must be accurate (reviewers check claims against the code)
+#   * no marketing words; the sentence ends with a period
+#   * do not hand-write an `npm:` key — npm linkage is picked up automatically
+#   * `tarball:` is recommended when the plugin is not on npm; the URL must be an
+#     https .tgz on GitHub Release hosting and the asset name stays version-free
+#     so `latest/download/` does not rot on the next release
+#   * CI also checks: repo declares dsh.bundle, repo is at least 1 day old,
+#     awesome-lint passes, and at most 3 entries per PR
+
+url: https://github.com/Ylight-cpu/dsh-creative-studio
+name: Ylight-cpu/dsh-creative-studio
+category: skill
+description:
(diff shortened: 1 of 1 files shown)


Ylight-cpu:
Updated the entry for **v0.2.0**, which adds four video effect capability groups (all covered by the end-to-end selftest, now 26 checks):

- `video-matte` — AI subject matting via RVM ONNX (auto-downloaded, ~14 MB) with background replacement, transparent-WebM and matte output modes, plus a chroma-key path for green-screen footage. Measured 81 fps at 568x320 and 17 fps at 720x1280 on CPU.
- `video-text-anim` — animated titles emitted as standard ASS: fade, four-direction slides, typewriter, pop, bounce, karaoke word highlight, lower-third.
- `video-join` — all 53 xfade transitions enumerated from the bundled ffmpeg, plus stylized presets: flash, glitch, whip-pan, soft-zoom, film-burn.
- `video-fx` — 13 glow / bloom / leak / trail / grain / vignette / sharpen / grading presets, chainable, with external LUT support and before/after comparison frames.

The description in this entry was rewritten to match those commands exactly. Still out of scope and stated as such in the skill docs: motion tracking, 3D camera moves, particle systems and rotoscoping (After Effects territory).

PR #4932, opened 2026-09-12
Add TodayJin/dsh-project-memory (memory)

TodayJin:
Adds `TodayJin/dsh-trilogy` under **memory**. (The repo was renamed from `dsh-project-memory`; GitHub redirects the old URL, and the entry file points at the current name.)

- `package.json` declares `dsh.bundle: { patch: "./cordis.patch.yml" }`, so it installs with `dsh plugin add`.
- Not on npm; it installs from source and needs no build step (plain ESM host half + a hand-written classic-script client half), so no `tarball:` field.
- `dsh.client` targets `web`; the plugin also loads fine in headless/TUI profiles (the browser routes are registered only when a `webServer` service exists).

Description accuracy: three files (`PROJECT.md`, `DECISIONS.md`, `SESSIONS.md`), three model tools (`memory_checkpoint`, `memory_read`, `memory_search`), one settings section, one composer chip. The three memory files are shown read-only in the settings panel; the only file editable there is the `AGENTS.md` instruction block. No vector store, no embeddings, no model calls - nothing in the plugin calls an LLM.


TodayJin:
Closing and reopening from the same branch: the `check` failure on this head is a diff-base artifact, not a property of the submission.

`pr-check.yml` diffs against `github.event.pull_request.base.sha`, which is frozen at `6060b3c` — the base commit when this PR was opened. The job checks out the PR merge ref, which now carries the current `main`, 124 commits later, including main's own `README.md` / `README.zh.md` change (+98/-27). So `git diff 6060b3c...HEAD -- README.md README.zh.md` is non-empty even though this PR touches neither file (`compare` reports one file, +6/-0), and the run takes the `--check` branch and reports main's README as out of sync.

A re-run cannot clear it: it reuses the stored merge commit, which still holds the pre-fix entry file (the `"name" is required` failure from the earlier attempt). A fresh PR gets a fresh base, so the diff is just this PR's own file again.

This submission adds one entry, `data/plugins/TodayJin__dsh-trilogy.yml`, with the required `name` field and a description aligned with the plugin as it is now.

PR #4943, opened 2026-09-12
Update Mandarin715/dsh-autostart (dev): resident supervisor, accurate 0.2.0 description

Mandarin715:
Updates the existing entry for `Mandarin715/dsh-autostart` so the description matches what 0.2.0 does. Only `data/plugins/Mandarin715__dsh-autostart.yml` is touched; the READMEs regenerate on merge. Category unchanged (`dev`).

What changed in the plugin, all of it verifiable in the repository:

- The restart path is now a **resident supervisor** that outlives the DSH host. Before the host is allowed to exit, the plugin confirms a supervisor is alive and writes the restart request to disk, then reads it back; if either step fails it refuses the restart and leaves DSH running. So a failed restart can no longer leave DSH down.
- DSH is spawned **attached to the hidden console the supervisor owns**, so tool commands no longer create a Windows Terminal window per call.
- The supervisor is deliberately **not a watchdog**: it starts a replacement only for a restart the user asked for, and steps aside when DSH exits on its own.

`docs/ACCEPTANCE.md` records the real-machine acceptance run (window enumeration, process lineage, `service.log` sequences). Release: https://github.com/Mandarin715/dsh-autostart/releases/tag/v0.2.0

### DIFF
--- a/data/plugins/Mandarin715__dsh-autostart.yml
+++ b/data/plugins/Mandarin715__dsh-autostart.yml
@@ -1,0 +1,6 @@
+url: https://github.com/Mandarin715/dsh-autostart
+name: Mandarin715/dsh-autostart
+category: dev
+description:
+  en: 'Windows boot autostart and one-click restart for the DSH service, from inside DSH: a resident supervisor that outlives the host, so a failed restart cannot leave DSH down; DSH spawned attached to a hidden console the supervisor owns, so no console window appears; plus a restart button, the current tokenised access URL and an optional post-start hook.'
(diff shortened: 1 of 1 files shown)


Mandarin715:
Closing this: the branch was created from a stale snapshot of my fork, so the entry file is added rather than modified and the PR conflicts with `main`.

I will re-open a clean one-file update after syncing my fork (the API sync is blocked for me because it touches a workflow file and my token has no `workflow` scope).

Sorry for the noise.

PR #4964, opened 2026-09-12
Refresh dsh-backup entry description (v0.10/0.11: typed backups)

xiaoyuyu6420:
Updates `data/plugins/xiaoyuyu6420__dsh-backup.yml` only — the entry description now mentions the v0.10+ feature set in addition to what was already there:

- typed backups via `/backup --types` with per-type rotation and subset merge-restore（分类型备份与按子集合并恢复）

Unchanged: url / name / category. `node scripts/generate-readme.mjs` runs clean locally with this file (generated README changes intentionally not committed — happy to include them if the gate prefers that shape).


fkysly:
Superseded by #4902, which you opened for the same entry and which covers a wider surface (typed backups **plus** `/backup migrate-check`, the credential sentinel, and the 0.12.0 wording). I verified 0.12.0 in the repo — `migrate-check`, the sentinel and `--types` are all there — and merged that one, so this diff no longer applies.

Nothing wrong with this PR; it just lost the race to its own sibling. Closing to keep one description per entry.

PR #4986, opened 2026-09-13
Add icanfinish11/dsh_superpowers

icanfinish11:
- [x] I added **one file** at `data/plugins/icanfinish11__dsh_superpowers.yml` — that single file is the whole submission. The READMEs are regenerated on `main` after merge: don't edit them by hand, and you don't need to commit them either / 我新增了一个 `data/plugins/icanfinish11__dsh_superpowers.yml` 文件——**这一个文件就是全部投稿**。README 会在合并后由 `main` 自动重新生成：不要手工编辑，也不需要提交
- [x] My repo's `package.json` declares **`dsh.bundle`** (not just `dsh.client`) / 仓库 `package.json` 已声明 `dsh.bundle`（只有 `dsh.client` 无法安装）— `"dsh": { "bundle": { "patch": "./cordis.patch.yml" } }`
- [x] My repo is at least **1 day old** / 仓库创建满 1 天 — created `2026-09-11T00:55:04Z`, ~53 h old at submission
- [x] `category` is one of the list / `category` 取值正确 — `skill`
- [x] Description states what the plugin does, no superlatives / 描述只说功能，不带营销词
- [x] My repo has the `dsh-plugin` topic / 仓库已打 `dsh-plugin` topic

## What it does / 插件做什么

**Adds the Superpowers skills (brainstorming, planning, TDD, debugging) to dsh, where they trigger automatically.**

- The 14 [Superpowers](https://github.com/obra/superpowers) skills are registered as a dsh skill catalog (`ctx.skills.registerProvider`), so the agent loads them with dsh's native `skill` tool. Each skill keeps its own directory as its resource base, so the scripts, prompts and templates a skill refers to resolve to their real location.
- The `using-superpowers` bootstrap is contributed as an always-on system-prompt section (`ctx.systemPrompt.section`, order 121). dsh has no session-start hook channel, so this is the always-on surface the porting guide's rule needs; unlike an injected history message it also cannot be lost to compaction.
- `references/dsh-tools.md` translates Superpowers' action vocabulary into dsh tools (`skill`, `todo_write`, `subagent`/`subagent_fork`, `pwsh`/`bash`, `grep`/`glob`, `web_search`, `ask_user_question`, …), and the same mapping is inlined into the bootstrap, so the model needs no extra file read.

## Claims ↔ evidence / 描述与代码的对应

| Claim in the description | Where it is true |
|---|---|
| the skills are added to dsh | `index.js` (provider over the bundled `skills/`); `tests/test-plugin.mjs` → "the bundled skills root resolves without any config" asserts all 14 names, their descriptions, `resourceBase` and rank |
| they trigger automatically | acceptance run on a fresh profile: with the prompt `Let's make a react todo list`, the session's **first tool call** is `skill({"name": "brainstorming"})`, before any file or shell action; the injected `<available_skills>` catalog lists them and the request header carries the `superpowers:using-superpowers bootstrap for dsh` section |

## Notes for review / 评审备注

- **No dependencies, no build step.** The plugin imports no packages at all — dsh's services arrive through `inject: ['skills', 'systemPrompt']` and everything else is a Node builtin — so there is nothing to declare as `peerDependencies` and no `allowBuilds` approval is triggered.
- **The skills tree is bundled on purpose.** `skills/` is upstream Superpowers (MIT, © 2025 Jesse Vincent) copied at `v6.3.0`, plus this adapter's own `references/dsh-tools.md` and one pointer line in `skills/using-superpowers/SKILL.md` — the single `SKILL.md` edit the Superpowers porting guide allows for a new harness. `LICENSE.superpowers` and `references/CREDITS.md` record the provenance, and `npm run sync-skills` refreshes the tree against a newer upstream checkout. A `github:obra/superpowers` dependency is not an option: pnpm rejects it as an exotic subdependency (`ERR_PNPM_EXOTIC_SUBDEP`), which would make the plugin uninstallable for everyone.
- Verified against a real harness, not just unit tests: `dsh plugin --profile headless add github:icanfinish11/dsh_superpowers` → the package is appended to `dsh.profile.bundles`, `dsh --profile headless --dump-config` shows the mounted row (`- id: superpowers`, `name: dsh-superpowers`), the smoke question "do you have superpowers?" is answered with the skill names, and the acceptance prompt above triggers `brainstorming`.
- `node tests/test-plugin.mjs` → 15/15 pass.


fkysly:
Read the whole thing, and the engineering here is better than the entry lets on: `upstream.tag` pinned at `v6.3.0` with `npm run sync-skills` reporting drift against a newer checkout, MIT + Jesse Vincent's copyright carried in `LICENSE.superpowers`, `references/CREDITS.md`, and a documented reason for vendoring rather than depending (`ERR_PNPM_EXOTIC_SUBDEP` would break every install). That is the honest way to ship someone else's tree.

The problem is that it is the **fifth** obra/superpowers port on the list:

| entry | what its line claims |
|---|---|
| `dennisrongo/dsh-plugins#dsh-superpowers` | catalog + compaction-safe bootstrap injection |
| `Meteor-system/superpowers-for-dsh` | 14 skills, agent preset, Windows sandbox compat |
| `JasonFreeLab/dsh-superpowers` | 14 skills as native DSH skills |
| `Wenaixi/dsh-superpower` | native SkillProvider (rank 550), Chinese docs, zero-build |

[contributing.md §4](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/blob/main/contributing.md): *whoever got here first keeps the slot — but that is a tiebreaker, not tenure … The rule is not first-come; the rule is whichever is better.* So this is answerable, but your entry line does not currently answer it. It reads *"Adds the Superpowers skills … to dsh, where they trigger automatically"* — which is what all four of the others also say, and it is the only one of the five that does not name **obra/superpowers** at all.

I will say plainly that the bar was applied loosely to entries two through four; that is our inconsistency, not a reason to hold you to a stricter one. But five rows pointing at one upstream is exactly the double-counting §4 exists to prevent, so I am not adding a fifth on a line that does not distinguish itself.

**What would change my mind, and it is a one-line edit:** the pinned-tag-plus-drift-report provenance is a real differentiator — none of the other four tell a reader *which* upstream revision they are running, and a skills tree that silently diverges from upstream is the failure mode here. Rewrite the description around that (name obra/superpowers, say the tree is pinned to a recorded tag and re-syncable) and I will take it.

If instead you think one of the four above is simply worse than yours and should lose the slot, say which and why — §4 allows that argument and I will read it. Marking `held: author`.

icanfinish11:
Thanks for reading it that closely. The line is changed:

> `Bundles the 14 obra/superpowers skills for dsh from a recorded upstream tag (v6.3.0), re-syncable with drift reporting; the skills trigger automatically.`

While I was there I finished the provenance the line promises: `package.json` now records `upstream.tag` **and** `upstream.commit` (`b36e0829…`), and `sync-skills --check` fails when the checkout no longer matches that pin — so the recorded revision cannot rot silently.

I also rebuilt the branch on today's `main`. The red `check` on my previous push was not the entry: that branch was based on a `main` whose generated READMEs no longer matched `data/plugins/`, and the check regenerates them. It is re-running.

**One correction, because it is the basis of the decision.** I checked the other four before claiming the pin as mine, and "none of the other four tell a reader which upstream revision they are running" does not hold:

- `JasonFreeLab/dsh-superpowers` — README: *"Ported from upstream \[obra/superpowers\] (v6.3.0, by Jesse Vincent / Prime Radiant)."*
- `Wenaixi/dsh-superpower` — README: *"上游基准锁定 `obra/superpowers v6.3.0`"* (upstream baseline locked at v6.3.0).
- `Meteor-system/superpowers-for-dsh` — `NOTICE.md`: *"vendors the skill content from \[obra/superpowers\], release 6.3.0, commit b36e082"* — that is more precise than mine, and it is the same commit this repo bundles.
- `dennisrongo/dsh-plugins` → `plugins/dsh-superpowers` — README describes serving a **pinned** copy when there is no clone to read instead.

So the pinned tag is not a differentiator, and I would rather say that than take the row on a premise I can show is wrong — an inaccurate line is the one thing your rules send back, and it would be inaccurate in a way a reader could check in five minutes.

What is actually mine, stated without comparison to the other four:

- the pin is machine-readable metadata (`upstream: { name, repository, tag, commit, synced }`), with a test asserting its shape so it cannot quietly disappear;
- `--check` reports drift per file and exits non-zero, and also fails on a pin/checkout mismatch;
- `LICENSE.superpowers` (upstream MIT verbatim) and `references/CREDITS.md` carry the attribution;
- the adapter has no dependencies at all and no build step.

If, on those facts, this is still a fifth row for one upstream that does not clear §4, close it and I will not argue — five rows pointing at one tree is what §4 exists to prevent, and I would rather the list stay accurate than gain a row. I am not proposing that any of the four lose its slot: I have not read their implementations closely enough to argue that, and would not want to pretend otherwise.

Disclosure, since it is relevant to how this was produced: the work was done with DeepSeek Harness (dsh) as an agent-assisted workflow, with this plugin installed in it. Every quote above is a line in one of the repositories linked here.

PR #4741, opened 2026-09-09
Add chromoany/dsh-notify-me (notify)

chromoany:
Adds one entry file: `data/plugins/chromoany__dsh-notify-me.yml`.

- repo declares `dsh.bundle.patch` → `./cordis.patch.yml` (committed at root) and `dsh.client.platform: web`
- `dsh-plugin` topic set; repo created 2026-09-08, npm package `dsh-notify-me@1.1.2` published

What it does: web client plugin — desktop notification, sound and tab-title marker when the agent needs input (approval, plan review, question) or a reply finishes in the background; toggled from the Settings page with zh/en notification language.

PR #5030, opened 2026-09-13
chore(dsh-notify-me): expand description keywords so it is findable in the market

chromoany:
### What

Rewrites `description.en` / `description.zh` for my own entry, `chromoany/dsh-notify-me` — no other entry touched.

### Why

The market search matches the description, and the current wording contains none of the queries a user would type. Searches that find nothing today: `提醒` · `消息提醒` · `完成提醒` · `后台完成提醒` · `message alert`. The new lines lead with those words in both locales, then keep the factual detail.

### Coverage

| query | before | after |
| --- | --- | --- |
| 提醒 / 消息提醒 / 完成提醒 / 桌面通知 | miss | hit |
| reminder / notification / message alert / alert | partial (`notification` only) | hit |
| 审批 / 提问 / 方案确认 / 提示音 / 音量 / 通知语言 | miss | hit |

### Accuracy

Every claim maps to shipped behaviour: pending `approval` / `plan-review` / `question` alerts, background turn completion, system toast + sound + `🔔 需要你` tab-title marker, and a Settings section with the master switch, sound/volume, per-case visibility toggles and a zh/en notification language. No numbers, commands or API names are asserted.

The READMEs are generated, so they are left alone and pick the new text up on regeneration after merge.


chromoany:
Closing this in favour of a delete-then-re-add pair. This PR touches the entry as a **modification**,
which the submission gate routes to maintainer review — correctly, per `contributing.md`. Our entry has
been queued that way since PR #4741 (open 2026-09-11 → merged 2026-09-13) and the marketplace search
still cannot find the plugin under `提醒` / `消息提醒` / `完成提醒`, since the description is the only
field that search reads.

So: this PR is closed, a one-file deletion PR follows, then a one-file addition PR carrying the new
description — the addition is the path CI squash-merges automatically. No other entry is touched in any
of the three, and the addition PR carries the same wording shown here.

Sorry for the extra queue traffic.

PR #5051, opened 2026-09-14
Add wangzhanchao883/dsh-hold-to-talk (voice)

wangzhanchao883:
# PR: add wangzhanchao883/dsh-hold-to-talk

## What this adds

One plugin entry to the curated awesome-dsh-plugin registry (this is what makes
it appear in the dsh-market plugin store):

- `data/plugins/wangzhanchao883__dsh-hold-to-talk.yml`

## Plugin

`dsh-hold-to-talk` — hold-to-talk speech-to-text in the DeepSeek Harness web
composer. Press and hold the mouse on the input box, speak, and the transcript is
inserted into the draft on release; sliding up cancels. A live transcript is
rendered in an overlay while holding, and recognition runs locally (SenseVoice via
sherpa-onnx), so there is no API key and no audio leaves the machine.

- Installable: declares `dsh.bundle` (`cordis.patch.yml`) + `dsh.client` (browser
  overlay and gesture handling). ✓
- Topic `dsh-plugin` set on the repo. ✓
- MIT license. ✓
- Repo created 2026-09-11T11:39:06Z, 13 commits, CI green on Node 22.x and 24.x.

This PR adds the entry file only; the READMEs are left to the post-merge
regeneration on `main`, as `contributing.md` asks.

## How it differs from the `voice` entries already on the list

The list already carries several composer voice-input plugins, so here is what
this one adds rather than repeats:

- **The gesture, not a button.** There is no mic button to click. Recognition is
  bound to a *mouse press-and-hold on the composer itself* (400ms threshold), the
  way WeChat desktop does it: release inserts the transcript into the draft, and
  sliding up more than 60px cancels. `haoku123/dsh-voice` (also host-side
  SenseVoice via sherpa-onnx) is a full-duplex voice mode driven by a key/tap
  toggle with spoken replies; `baisama-cloud/dsh-stt-input`, `NewDaNew/dsh-voice-input`
  and `0nt-one/dsh-voice-input` are mic-button dictation. None of them use the
  hold-the-mouse-on-the-input-box gesture or slide-up-to-cancel.
- **Partial text is never written back.** Interim results are painted into a
  floating overlay only; the composer draft is touched once, on release, from the
  complete audio — so a cancelled hold leaves no half sentence behind.
- **Fully offline, no key, no cloud fallback.** SenseVoice int8 via sherpa-onnx,
  decoding on a worker thread (RTF ≈ 0.15 native). The 228MB model is fetched
  once on first use from a configurable mirror and is not in the repo.

If the maintainers judge it covered by an existing entry, say so and I will not
argue the slot — but the interaction itself is not on the list yet.

## Notes for review

- Category is `voice` — speech input for the composer. If a better fit exists,
  please re-file; no objection.
- `npm test` runs without DSH, without a browser and **without the 228MB model**:
  the suites that need the model skip themselves and the model-free subset runs
  for real, so a green CI run on a clean machine is expected.
- The SenseVoice int8 model (228MB) is downloaded once on first use from a
  configurable mirror (`hf-mirror.com` by default). The repo ships no model.
- `@deepseek-ai/schemastery` is declared as an **optional** peerDependency, and the
  settings namespace registers through a dynamic import that degrades gracefully
  when the package is absent (the core features do not depend on it). The only
  runtime dependencies are `sherpa-onnx-node` (native, per-platform optional deps)
  and `sherpa-onnx` (WASM fallback).
- No credentials are read or stored. Audio travels only over loopback HTTP to the
  plugin's own host route, which refuses cross-site requests; the microphone opens
  only while the mouse is held.
- `tools/zh.wav` is a 5.6 s test clip from the k2-fsa sherpa-onnx examples
  (Apache-2.0), included so the decode smoke test works on a fresh clone.
- Screenshots are declared in the plugin repo root as `screenshots.json` (three
  real captures plus a demo GIF under `assets/screenshots/`), served from
  `raw.githubusercontent.com` — verified reachable.


wangzhanchao883:
To be upfront about the overlap first: `haoku123/dsh-voice` is on the list
already, and it does offer a hold-to-talk dictation mode with a live caption. I
am not going to argue that this entry covers different ground — it partly does.
The one thing it does differently is the reason the plugin exists, so let me put
that on the table instead.

**The gesture is mouse-only, start to finish, and it lives on the composer.**

Hold the mouse on the input box, still, for 400 ms: recording starts. Speak.
Release: the transcript is appended to the draft. Slide up more than 60 px while
holding: the overlay turns red and the audio is discarded.

There is no keyboard in that loop, and no icon to aim at. That is the whole
point. The workflow this was built from is typing with one hand while the other
hand is doing something else — a child's homework, a soldering iron, a phone
call — and reaching for `Ctrl` or a mic button is exactly what breaks it. With
this one you can start, speak, and land the text without moving your hand off
the input area and without a second hand involved at all.

Three details that fall out of that choice, in case they matter to the review:

- **Nothing is written until you release.** The overlay transcript is a preview
  from a re-decoded tail window; the text that goes into the draft is decoded
  from the full audio at release. So a hold you cancel leaves the draft exactly
  as it was.
- **It does not disturb normal editing.** A quick click still places the caret,
  dragging still selects, moving more than 8 px (or making a selection) during
  the press is treated as ordinary mouse work and records nothing. `Ctrl` is
  never captured.
- **`Esc` also cancels**, for the cases where a hand does happen to be on the
  keyboard.

Recognition is the same local path — SenseVoice int8 via sherpa-onnx in a worker
thread, no API key, audio never leaves the machine. The published package is
`dsh-hold-to-talk@0.1.2` (MIT, `dsh.bundle` declared): 9 files, 86 KB unpacked.
The model is the only heavy part, and it is downloaded on first use, not
bundled.

So: is the mouse-only, composer-resident hold worth a row of its own next to
`dsh-voice`, or do you read the two as covered? If it is the latter, say so
plainly and I will close this without arguing the slot — I would rather know
than leave it sitting here.

wangzhanchao883:
Superseded by a rebuild from the current `main` (the entry is unchanged apart from the description, which now leads with the interaction instead of the recognizer).

Closing this one rather than bumping it: it has been open since 09-14 and the entry itself is still what this plugin should ship as — it just needs to enter the queue as a fresh PR. Nothing here needs a reply.

PR #5078, opened 2026-09-14
Add igormel81/dsh-chat-cost

igormel81:
Adds one entry file, no generated README touched.

- repo: https://github.com/igormel81/dsh-chat-cost
- entry: data/plugins/igormel81__dsh-chat-cost.yml
- category: usage
- package: dsh-chat-cost on npm, installable with `dsh plugin --profile web add dsh-chat-cost`
- manifest: package.json declares dsh.bundle.patch -> ./cordis.patch.yml
- repo age: created 2026-09-12T18:14:28Z

What it does, in one line: live token cost for a chat, its subagents and the whole
session tree, the price of each finished answer in the transcript, budget-aware
work plans packed under a money limit, and an append-only JSONL cost log written
into the project folder.

Everything the description claims is checkable in the repository: the seven
providers and 139 priced models are in data/prices.json, the six tools are in
lib/tools.js, the per-turn pricing is lib/turns.js, and the test suite is 140
tests (`npm test`). Limits are stated in the README too: a chat the log has never
seen and that is not open shows a dash rather than an estimate, and long-context
tier prices for some Gemini and Grok models are not applied yet.


igormel81:
Since the description is read as a claim about the code, I added a map so the check takes minutes rather than an afternoon: [`submission/awesome-dsh-plugin/claims.md`](https://github.com/igormel81/dsh-chat-cost/blob/main/submission/awesome-dsh-plugin/claims.md) lists every sentence of the entry with the path that proves it — the catalog file for the seven providers and 139 priced models, `lib/turns.js` for the per-answer price, `lib/scenarios.js` for the adequacy rule, `lib/log.js` for the append-only ledger and the `.gitignore` behaviour, and the six tools in `lib/tools.js`.

Two things I would rather state than have found:

- The catalog holds 139 priced models, but only 19 of them publish **both** a cache read and a cache write rate (80 publish a read rate, 19 a write rate). The entry says "wherever the provider publishes them" rather than implying all 139.
- Long-context tier prices published for 22 catalog models are **not** applied yet; the base tier is used. `tiers` in `data/prices.json` is read by nothing in `lib/`. That is in the README's limitations too.

Nothing in the repository changed in a way that affects the submitted file — this is the same `data/plugins/igormel81__dsh-chat-cost.yml`, unchanged since the branch was pushed.

PR #5125, opened 2026-09-15
Add xie129716/computer-user-vision

xie129716:
Adds one entry for `xie129716/computer-user-vision` under `tools`. It is an unofficial fork of
[jing-hy/computer-user](https://github.com/jing-hy/computer-user), **which is already on this list**
(`data/plugins/jing-hy__computer-user.yml`, also under `tools`). Please read the next section first —
that overlap is the decision this PR is asking for, not a detail underneath it.

## The slot question, first

This is not an addition beside that entry: the two cannot be installed together — same package name,
same cordis row id, stated just below — so one of them has to be the row. The review rules say a fork
*is* added when it is the better-kept one or when it genuinely adds something, and that "the rule is
not first-come; the rule is whichever is better". I am not asking you to take that on faith.

**Upstream does not load on the DSH that ships today.** Its `src/index.js` opens with

    import { settingsNamespace } from '@deepseek-ai/dsh-settings';

and `@deepseek-ai/dsh-settings@0.1.5-rc.2` exports exactly four names — `SettingsConflictError`,
`SettingsProvider`, `default`, `redactSecrets`. A missing named export fails the whole ES module at
link time, so `apply()` never runs: the plugin appears in `dsh --dump-config` and then does nothing,
with nothing in the boot log. Both halves are reproducible in about a minute:

    node -e "import('@deepseek-ai/dsh-settings').then(m=>console.log(Object.keys(m).sort().join(', ')))"
    # SettingsConflictError, SettingsProvider, default, redactSecrets   <- no settingsNamespace

    curl -s https://raw.githubusercontent.com/jing-hy/computer-user/main/src/index.js | head -3

The fork's fix is `import * as settingsModule from '@deepseek-ai/dsh-settings'` — a namespace import
cannot fail at link time on a missing name — commented at `src/index.js:63`, with the mechanism
written up in `docs/adaptation-notes.md`.

Maintenance points the same direction: upstream last pushed 2026-08-27; this fork 2026-09-14, 18
releases, 0.3.32 (upstream's npm tag is 0.3.6).

So the outcome I am asking for is either "list this one, and drop or annotate
`jing-hy__computer-user.yml`", or whichever form you would rather have. I maintain the fork, not
upstream, and the call is yours to make on the evidence above.

Why it is a binary choice rather than an addition: the package name is still `computer-user`, and
`cordis.patch.yml` registers cordis row id `computer-user` — deliberately, so a profile can drop this
fork in exactly where the original sat. A profile cannot hold both: one `node_modules/computer-user`,
one row id. Renaming would break the drop-in, which is why it is stated here rather than changed
quietly.

## What the fork adds, all of it in the repository

- **Ported to the current DSH settings API (the 0.1.5 line), without dropping the older one.** 0.1.5
  removed the module-level `settingsNamespace()` helper upstream imported by name. The fork imports
  the module namespace and feature-detects it —
  `settingsModule.settingsNamespace?.(value) ?? value` at `src/index.js:73` — registers with
  `provider.register(namespace, Config, { base })`, and writes through `SettingsScope.update(patch)`,
  the 0.1.5 write path (there is no `scope.set`). It runs on 0.1.5 and still runs on the older line.
- **The screenshot goes to the model as an image — no external OCR step.** The tool asks the harness
  what the routed model accepts (`ctx.llm.resolveModelInfo(provider, model)`, then
  `inputModalities.includes('image')`) instead of trusting a manual switch or a guess; for
  `deepseek-flash` (DeepSeek-V41-Flash) that declaration includes `image`. On such a route the
  capture is saved through the host's `attachments.saveImage()` and rides the tool result as a real
  image block; on a text-only route it falls back to the PNG path and reports `vision: false`.
  Upstream had to be looked at through an external OCR tool because the model could not see the
  screen; this fork does not, so `picturereader` becomes a fallback rather than a requirement.
- **The coordinate mapping stays exact through the host's own downscale.** The capture is fitted to the
  adapter's declared `imagePixelBudget` (640,000 — the same `DEFAULT_REQUEST_IMAGE_PIXEL_BUDGET`
  `dsh-llm-deepseek` sets for this model), so the server does not downscale a second time.
  `screen_per_pixel` then folds the capture's own crop/scale together with whatever normalization the
  host applied when saving, and the envelope prints the resulting formula for the model:
  `screen_mapping: screen_x = vx + image_x * kx ; screen_y = vy + image_y * ky`.
- **Element refs, so a click stops depending on a pixel estimate.** A 1920x1080 desktop is
  2,073,600 px, above the 640,000 px vision budget, so the preview a model reasons about is
  ~1045x588 — one image pixel is 1.84 screen pixels and a 22 px toolbar button is 12 px tall in
  that picture. `computer_screenshot` and the new `computer_elements` now enumerate the focused
  window's UI Automation controls and return refs; `computer_click` takes `ref` or `name` and
  the OS supplies the exact rectangle, so the measured click error is 0 px. Enumeration tests
  capability rather than control type, because a WinForms button reports as `ControlType.Pane`
  with no patterns at all.
- **Exact geometry, and honest geometry.** `computer_list_windows` returns the DWM extended frame
  bounds as `rect` — `GetWindowRect` is 8 px larger on every side (it includes the invisible
  resize border), which made every window-relative aim wrong by 8 px. Both values are reported.
- **One process per tool call.** Each PowerShell start costs ~380 ms of process creation plus
  `Add-Type` compilation, and a click used to spawn three or four of them (~1.1-1.5 s). The three
  one-shot scripts are merged into `src/act.ps1`; the focus check, ref resolution, the click and
  the probe now happen in the same process. Moves also use absolute `SendInput` over the virtual
  desktop and are verified before the click follows.
- **PerMonitorV2 DPI awareness.** `powershell.exe` starts DPI-*unaware*, so on a scaled display
  Windows virtualised every coordinate; the old scripts only reached System-aware.
- **Pre-flight focus guard.** Every input tool takes an optional `expect_window`; on a mismatch
  it refuses *without sending any input* and returns the window it found. Reporting
  `focused_window` only *after* typing is a post-mortem — during acceptance a `Ctrl+H` meant for
  Notepad landed in the browser that way.
- **A control indicator you can see, and stop.** While the AI holds control, `src/overlay.ps1` draws a
  slowly pulsing gradient frame on all four screen edges, a halo that follows the mouse and changes
  shape with the cursor, and a top banner naming the controller with a Stop button and a
  `Ctrl+Alt+Esc` global hotkey (`MOD_ALT|MOD_CONTROL` + `VK_ESCAPE`). The `stop` marker is written
  **only** by the user; while it exists every `computer_*` call is refused, and it stays down until
  the user re-approves — by `/computer`, the switch below, or merely sending a new message. The idle
  indicator auto-hides after 25 s; a user stop does not.
- **A computer-use switch in the chat input.** Registered at `conversation.input.left`, beside the
  composer, with three visible states: on, off, and *stopped by you* (amber). Turning it on grants
  approval silently — exactly what `/computer` does — and turning it off stops control immediately and
  blocks further calls. The full settings page registers a `settings.section` of its own.
- **A mode gate and disk-backed approval.** `disabled / readonly / manual / auto`, plus
  `approval_scope: session | profile` — session keeps one `/computer` valid for that session, profile
  makes it valid for every session until revoked. Both go to disk, so they survive a host restart, and
  `ai_can_change_mode` defaults to **off**, so the model cannot widen its own permissions.

## Two things to know before listing

**On "do its dependencies point at the original".** That rule governs bundles, and this is not one: it
ships behaviour, and it has **no `dependencies` at all** — the only entries are `peerDependencies` on
the harness's own `@deepseek-ai/*` packages, with an explicit prerelease branch per tuple:
`>=0.1.0-rc.6 <0.1.5-0 || >=0.1.5-rc.1 <0.2.0-0 || >=0.2.0-rc.1 <0.3.0-0`. Nothing here resolves to a
copy of anyone's work.

**The `tarball:` field is load-bearing, not cosmetic.** npm's `computer-user` is upstream's package
(`repository` -> `jing-hy/computer-user`), so anyone installing by that name gets upstream's code —
the version that does not load. The release tarball is the only install path for this fork, so please
keep the field. The corollary is expected rather than a defect: the npm-linkage rule will not attach a
download figure here, because npm's `computer-user` points at a different repository.

## Flagged rather than left to be found

- This repository is **not** a GitHub fork (`fork: false`, no `parent`). It was built from the
  published 0.3.6 tarball rather than from a clone, and GitHub sets fork status only at creation, so
  it cannot be added afterwards. The relationship is recorded in `package.json` instead: `author`,
  an explicit `forkedFrom`, and a `repository` pointing at this repository, with upstream's
  copyright line kept in `LICENSE` (MIT).
- `screenshots.json` declares 4 relative paths; all exist and stay inside the plugin directory, and
  they are pictures of the two user-visible controls above — `control-indicator.png` and
  `stop-banner.png` for the overlay, `switch-on.png` and `switch-stopped.png` for the chat switch.

## Not a duplicate of the other computer-use entries

`988hj7tczd-oss/dsh-computer-use`, `Anionex/dsh-computer-use`,
`qphotoai/dsh-computer-use-windows`, `Yu-tao-Li/dsh-computer-use-win`, `ZRui-C/dsh-computer-use`,
`JohnXu22786/computer-control` and `Fish121380/auto-mouse` are separate implementations; this one is
PowerShell/SendInput with UI Automation refs and a user-visible stop.

## Checks I ran before submitting

- `dsh.bundle` is declared in the root `package.json` with a `cordis.patch.yml` beside it.
- `scripts/check-submission.mjs --only-list` against current `main` — **pass**.
- `scripts/generate-readme.mjs` — exit 0, and the entry line renders.
- `scripts/check-bleed.mjs` — **no pair**: this description shares no 40-character run with any
  existing entry, including the upstream one it forks.
- The description is 304 characters (the list runs 182 at the median and 314 at p90), so there are
  fewer claims to check; every remaining one is countable or greppable.
- `verify/vision-screenshot.mjs` — **16/16**, including that an image block is attached on an
  image-capable route, that none is attached on a text-only route, and that `screen_per_pixel` still
  reproduces the real screen size after the host halves the picture
  (522 * 3.6782 = 1920.0 vs screen 1920).
- `releases/latest/download/computer-user.tgz` returns HTTP 200 and the asset name is version-free,
  so it will not 404 on the next release.
- Official packages are `peerDependencies`, with an explicit prerelease branch per tuple. The
  previous `>=0.1.0-rc.6` silently excluded `0.1.5-rc.2`, the build actually shipping — verified
  with node-semver, not by eye.
- Installed into a real `web` profile and exercised: `apply()` registers 13 tools, the settings
  namespace resolves through the 0.1.5 provider API, and `/computer` registers.


fkysly:
合了。分叉类投稿我按 [contributing.md §4](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/blob/main/contributing.md) 多看了几眼，结论是它站得住，把依据写在这里，省得以后有人问。

原作 `jing-hy/computer-user` 已经在列表里（9 个工具，最后推送 2026-08-27）。规则原文是「一个分叉**是可以**被收录的——只要它维护得更好，或者确实做了新东西。规则不是先来后到，规则是谁更好」。你这边两条都占：

- **确实做了新东西**：13 个 `computer_*`（原作 9 个），而且图像直出 + 图像→屏幕精确映射把原作「配合 picturereader 让纯文本模型跑 OCR」那条链路直接省掉了；控件以 UI Automation 引用返回也是原作没有的。
- **维护得更活**：今天还在推送，原作停在三周前。

**署名也到位**，这点我专门查了：你不是 GitHub fork（`fork: false`，是手工复制），所以没有自动继承的 fork 关系和署名，但 README 里点名了 `jing-hy/computer-user` 并标了 Upstream，`LICENSE` 也在。描述第一句就写 "forked from computer-user"——先说清楚来历再说自己加了什么，这是对的做法。

`Ctrl+Alt+Esc` 阻断所有调用那一条，是我读下来最想有的功能。

PR #5137, opened 2026-09-15
Add Neptune810/dsh-model-router

Neptune810:
- [x] I added **one file** at `data/plugins/Neptune810__dsh-model-router.yml` — that single file is the whole submission
- [x] My repo's `package.json` declares **`dsh.bundle`** (not just `dsh.client`)
- [x] My repo is at least **1 day old**
- [x] `category` is `model`
- [x] My repo has the `dsh-plugin` topic

**What it does.** Host-only plugin, no browser UI. It classifies each step and sets the reasoning effort of the DeepSeek flash model to match — `trivial` → `off`, `standard` → `low`, `engineering` → `high` — and only steps the class up on tool-failure evidence (failing tool results, or the same call retried with identical arguments). The model itself never changes, and `max` is opt-in.

**Where the description's claims come from:** `trivial` → `off` / `standard` → `low` / `engineering` → `high` are the shipped route table in `lib/policy.js`; "raises it only on repeated tool failures" is the absence of a ratchet (`scoring.turnPerPoint: 0`, tool calls counted per task) plus `escalateOnErrors` / `escalateOnRepeats`; "effort max is opt-in" is `allowMax: false` by default, clamped at the effort level so even a hand-written route table asking for `max` is demoted.

**Tests:** `node --test` — 36 passing. Repo declares `dsh.bundle.patch: ./cordis.patch.yml` at the root.


Neptune810:
Flagging the nearest existing entries myself, so the overlap question is on the record rather than found later — happy to be told this is already covered:

- [`Electricitysheep/dsh-tool-turbo`](https://github.com/Electricitysheep/dsh-tool-turbo) — per tool round, injects the *lowest sensible* effort based on the step's recent tool calls and payload sizes, to cut the thinking between tool calls.
- [`drscrewdriver/dsh-thinking-levels`](https://github.com/drscrewdriver/dsh-thinking-levels) — `Auto` in the session model selector, scheduling low/high/max from the recent tool-call history; also exposes manual wire levels and adds selector UI.

What this entry does differently, all of it in `lib/policy.js`:

- **It classifies the step's own shape, not the tool history.** The class comes from the claimed user message — a clearly cheap intent, engineering cues, code/diff/XML structure, a dense brief — and the shipped ladder is `trivial` → `off`, `standard` → `low`, `engineering` → `high`.
- **Nothing accumulates.** `scoring.turnPerPoint` is `0` and tool calls are counted per task, so a long agent run cannot drift upward. Effort rises only on failure evidence inside the current task — failing tool results (`escalateOnErrors: 2`) or the same call retried with identical arguments (`escalateOnRepeats: 3`) — capped at `maxEscalations: 2`.
- **`max` is opt-in and clamped at the effort level**, so a hand-written route table asking for `max` is demoted too, and a manually selected `max` is demoted as well on a managed model (`demoteManualMax`).
- **It never picks a model.** It drives `deepseek-flash` on the `deepseek-official` provider and only sets effort; there is no client bundle, so nothing appears in the browser UI.

If that still reads as covered by those two, that is a fair call to make.

PR #5149, opened 2026-09-15
fix(probe-updates): probe the least recently checked repos first

po-et:
Found while looking at dsh-market/dsh-market#614 (a plugin whose `updates.json` entry was stuck on a release from 2026-08-31).

The nightly `probe-updates.mjs` walks every listed repository in README order, two API calls each, and the token's hourly budget reaches about a third of them. The 2026-09-14 scheduled run logged:

```
3632 listed, 3632 to probe (PROBE_ALL)
2364 repo(s) failed the first pass — retrying serially
2364 repo(s) kept their previous update data
updates.json written: 3396 repos (1268 refreshed this run)
```

Because the order is the same every night, the same stretch is starved every night. In the published `updates.json` today, grouped by README position: the first 30% and the last 10% of the list were checked on or after 09-12, while positions 40–80% (about 1,300 entries) were last checked on 09-05 — the head gets through before the quota runs out, the tail after it resets during the serial retry, and the middle never does. `mafeis/dsh-net-proxy` sits at position 1828 of 3722 and has been on `checkedAt: 2026-09-05` with `v0.2.6` while GitHub is at `v0.7.19`.

This sorts the pending list by `checkedAt`, never-checked first, so whatever the budget does not reach today is first in line tomorrow; ties keep README order, so a fully fresh list still probes in list order. `PROBE_ALL` is unchanged — it still re-checks everything, just in a rotating order. Small helper in `scripts/lib/probe-order.mjs` with a `node --test` file, wired into the same PR-check step as `added-dates.test.mjs`.

Not touched here: `probe-readmes.mjs` / `probe-stars.mjs` have the same first-pass/retry shape and may benefit from the same ordering, but I have not measured them.


Yur0918:
Heads-up on your checks: `Repository config guard` is failing (10:23), and I
think the cause is the `.github/workflows/pr-check.yml` hunk in this branch
rather than anything in the comparator.

[`pr-guard.yml`](https://github.com/awesome-dsh-plugin/awesome-dsh-plugin/blob/main/.github/workflows/pr-guard.yml)
lists the PR's files and flags any path starting with `.github/`; its own header
says a PR that touches `.github/` "is not a reason to reject the submission,
only a reason it must not merge until those files are dropped from the branch".
It is complaining about `path: .github/workflows/pr-check.yml`.

That leaves a Catch-22 worth knowing about: the step you extended is
`node --test scripts/added-dates.test.mjs` — an explicit single-file list, no
glob — so wiring a new test file into CI requires editing that file, which the
guard then flags. Dropping just that hunk should clear the red; the test file
itself stays fine to keep, and a maintainer can add it to the step on merge.

Two things I can contribute, since I ended up auditing the same problem from the
market side (dsh-market/dsh-market#614):

1. **Independent measurement.** 09-14's run log shows what the budget did; I
   replayed both orderings over the published payload instead — 3,477 entries,
   1,177 of them still on `checkedAt: 2026-09-05`, against a 1,000-entry budget:
   README order reaches **1** of those 1,177, `checkedAt` ascending reaches
   **819**. Counterfactual replay over the published data, not a probe run — but
   it is the same conclusion from the other direction.
2. **A duplicate to ignore.** I opened #5170 for this before seeing yours. Same
   comparator, six hours later, and without the test file — so I am closing it
   in favour of this one. Nothing to reconcile; yours is the one to follow.

If it would help, I am happy to drop the workflow hunk onto this branch (or
rebase it) — say the word.

po-et:
Thanks — dropped the `.github/workflows/pr-check.yml` hunk, so the branch no longer touches `.github/`; `scripts/probe-order.test.mjs` stays and runs green with `node --test scripts/added-dates.test.mjs scripts/probe-order.test.mjs`, so it is one word to add to that step whenever a maintainer wants it in CI. And thanks for the replay over the published payload — 1 of 1,177 stale entries reached in README order against 819 with `checkedAt` ascending is the same shape the 09-14 log showed, from the other side.

po-et:
One more data point from the same 09-14 run log, for why it is this probe and not the others: the nightly runs `probe-npm`, `probe-downloads`, `probe-stars`, `probe-readmes` and then `probe-updates`, in that order, on one token. Stars refreshed 3,619 of 3,631 and downloads all 1,827; updates, last in line and two calls per repository, is the one that meets the exhausted budget (`2364 repo(s) kept their previous update data`). Ordering the pending list by `checkedAt` is enough to rotate it through; moving the step earlier would only shift which probe starves.

Yur0918:
Took a look at the two you flagged as unmeasured — an audit of all seven probes,
since "same shape" turns out to be four of them rather than two.

Four build their pending list with the identical un-sorted idiom:

- `scripts/probe-downloads.mjs:71` — `published.filter(([url]) => !fresh(map[url]))`
- `scripts/probe-stars.mjs:63` — `urls.filter((url) => !fresh(map[url]))`
- `scripts/probe-readmes.mjs:159` — `urls.filter((url) => !fresh(map[url]))`
- `scripts/probe-updates.mjs:139` — the one this PR fixes

`probe-npm`, `probe-tarballs` and `probe-screenshots` have no pending list, so
ordering does not apply to them.

Two things worth knowing before anyone generalizes the comparator:

**readmes timestamps are `fetchedAt`, not `checkedAt`.** `probe-readmes.mjs:156`
reads `entry.fetchedAt` where `probe-updates.mjs:137` reads `entry.checkedAt`. A
copied comparator asking for `map[url]?.checkedAt` gets `undefined` for every
readme entry, every key compares equal, and the list stays in README order — a
silent no-op rather than something that breaks loudly. Whoever ports it should
pass the field name in.

**readmes is the one sitting in the middle.** It is fourth of the seven, between
stars (third, which you measured at 3,619/3,631) and updates (fifth, which
starves), and it is the only one of the four with no committed floor:
`data/readmes.json` is gitignored and exists only in the Actions cache, while
`data/stars.json` and `data/downloads.json` are deliberately committed — the
`.gitignore` says so — precisely so that "a failed repo keeps its old entry"
stays true. There is nothing behind a starved readme fetch: the entry keeps its
last good README, or renders with none if it is new. Its own comment at lines
177-182 describes exactly that.

One caveat on the reason, though: readmes attributes its first-pass losses to
burst — "this pass issues roughly sixty requests a second, trips GitHub's
secondary limit" — not to the hourly budget, and it already retries serially at
250ms. So ordering would rotate *which* entries are at risk rather than reduce
the loss. Still worth doing, for your reason rather than the budget one:
whatever is not reached tonight is first tomorrow. Just a different starvation
than updates'.

Happy to send that change for downloads/stars/readmes if you would rather not —
say the word, or leave it if you are taking it.

fkysly:
合了，而且**我把你说的两个数字都验了**，都对得上：

```
mafeis/dsh-net-proxy
  updates.json: release v0.2.6 (2026-08-31), checkedAt 2026-09-05
  GitHub 实际:  v0.7.20 (2026-09-16)      ← 落后 5 个小版本、12 天
```

`updates.json` 里 `checkedAt` 的整体分布更说明问题：

| checkedAt | 条数 |
|---|---|
| 2026-09-05 | **1,177** |
| 2026-09-14 | 124 |
| 2026-09-15 | 461 |
| 2026-09-16 | 572 |
| 2026-09-17 | 1,186 |

**1,177 条（33%）整整 12 天没被探测过**，而且集中在同一段——正是你描述的「每天都从同一个地方开始，预算用完就停在同一个地方」。这不是数据错，是覆盖不均，而且不均的形状每晚一模一样。

**排序修复是对的**，我也跑了你的四个用例（4/4 通过），包括「不改动传入数组」和「无 checkedAt 视为从未检查」这两条边界。`--reverse` 之类的花招会让「平局保持 README 顺序」失效，你用稳定排序 + 纯函数做到了，比我预想的小。

**顺带说一下另一半**，因为你提到「probe-readmes / probe-stars 有同样的形状、但没测量」——我前天测了并且修了 readmes 那条：

- 根因是配额本身：GitHub Actions 的 token **每小时 1,000 次 REST**，而 `probe-readmes` 每仓库要 4 次（README API + 列目录 + 取文件），3,727 条就是上万次。两晚各失败 3,334/3,722（89.5%）。
- 修法是把文件内容改从 `raw.githubusercontent.com` 取（**完全不计入配额**），并用一次批量 GraphQL 拿目录列表把「靠猜文件名」变成「知道」。首轮失败 3,334 → **46**，抓取阶段 20 分钟 → 3 分 46 秒，写入条目 3,403 → 3,722（差的 319 条是**从来没有过 README** 的条目）。
- 实测：昨晚首次无人值守的 nightly 保持住了（46 条失败、8.5 分钟）。

**所以你的排序和这个修法是互补的两件事**：raw 把天花板拆掉，排序保证在还有天花板的地方（`updates` 仍在 REST 上，每仓库 2 次调用）不会反复饿死同一段。两个都必要，我把 `updates` 走 GraphQL 的可能性记在 #5275 里了——你这条先把「谁先谁后」的问题解决了，那是当下能立刻见效的一半。

谢谢，这份诊断的精确度（给出位置 1828 和具体的版本差）省了我重新测量一遍。

Issue #3512, opened 2026-08-28
dsh-plugin-catalog: npm publish stalled since 2026-08-26 — china-region market catalog frozen at 08-25 (2189 vs 2379 entries)


hi-fangj:
### What kind of problem? / 问题类型

Something else about the list / 其他清单相关问题

### Which entry? / 哪个条目

https://github.com/hi-fangj/dsh-models-radar — used only as a concrete example; the problem affects every entry added after 2026-08-25.

### What's wrong? / 具体问题

**Summary / 摘要**

The daily npm publish of `dsh-plugin-catalog` appears to have stopped. Its newest version is `2026.826.2432`, published **2026-08-26 03:47 UTC**, while the live `plugins.json` on awesome-dsh-plugin.com is already at `updated: 2026-08-27`. dsh-market's `china` download region reads this npm package (via npm mirror) as its **primary** catalog source and only falls back to the live URL when the fetch **fails** — so as long as the stale package keeps being served successfully, China-region users browse a catalog frozen at **2026-08-25 (2189 entries)** instead of the current **2379**.

`dsh-plugin-catalog` 的每日 npm 发布似乎停了：最新版 `2026.826.2432` 发布于 2026-08-26 03:47 UTC，而 awesome-dsh-plugin.com 上的 live `plugins.json` 已更新到 08-27。dsh-market 的 `china` 下载区域把该 npm 包作为**首选**目录源（仅当拉取**失败**才回退 live 地址），因此只要这个过期的包还能正常返回，中国区域用户看到的目录就停在 08-25（2189 条），而不是当前的 2379 条。

**Evidence / 证据（all measured 2026-08-28）**

| item | value |
|---|---|
| npm latest (registry.npmjs.org) | `2026.826.2432`, published `2026-08-26T03:47:19Z` |
| npm latest (mirrors.cloud.tencent.com) | same version — mirror is in sync, so this is **not** mirror lag |
| package content (`package/plugins.json`) | `updated: 2026-08-25`, `count: 2189` |
| live `https://awesome-dsh-plugin.com/plugins.json` | `updated: 2026-08-27`, `count: 2379` |
| concrete example | `hi-fangj/dsh-models-radar`, added 2026-08-27 — present in live, absent from the package |

**Why China users never fall back / 为什么中国区域不会自动回退**

[...]

Issue #1673, opened 2026-08-18
plugins.json stars field is null for ALL entries (2026-08-18) — probe-stars.mjs failure not falling back to committed data


oinzc:
## 问题描述 / Problem

`https://awesome-dsh-plugin.com/plugins.json` (v1.13.0, updated 2026-08-18) returns `"stars": null` for **all 1362 entries**. The dshmarket plugin market UI (which consumes this file) shows no star counts / ratings on any plugin.

数据源今天（2026-08-18）生成的 `plugins.json` 中，全部 1362 个插件的 `stars` 字段都是 `null`，导致 DSH 插件市场界面看不到任何 star 数量。本地 2026-08-17 的离线快照中 stars 字段是正常的（例如 `dsh-message-rail: 2`），说明是今天生成时才坏的。

## 复现 / Repro

```bash
curl -s https://awesome-dsh-plugin.com/plugins.json | jq '.[0]'
# → { "name": "dsh-status-rotator", ..., "stars": null, ... }
```

Count of entries with non-null stars: **0 / 1362**.

## 初步定位 / Suspected cause

`scripts/probe-stars.mjs` 的设计（源码注释）：

> "Without a token the script exits 0 without touching the file, so offline builds keep the last committed data. **A failed repo keeps its old entry.**"

失败时"保留旧值"的兜底依赖 `data/stars.json` 里的**已提交旧数据**。但今天的行为表明：

1. 要么 CI 运行时的 `GITHUB_TOKEN` 触发了 GitHub API 限流/超时（今天我本机直接访问 `api.github.com` 也遇到 `504 Gateway Timeout`），且
2. `build-site.mjs` 在 `probe-stars` 失败/产出空数据后，没有阻止 `stars: null` 写入 `plugins.json`——把空值直接发布了出去。

GitHub API 侧目前是正常的（直接查 `api.github.com/repos/01Virex/dsh-status-rotator` 返回 `stargazers_count: 21`），所以是生成链路的失败兜底没有生效。

## 建议修复 / Suggested fix

1. 在 `build-site.mjs` 中，若 `probe-stars` 产出全空 stars，则**拒绝发布**（保留上一版 plugins.json），或
2. `probe-stars.mjs` 失败时回退到 `data/stars.json` 的旧值而不是写 `null`（确保"失败保留旧值"的承诺真正生效），或
[...]

PR #5170, opened 2026-09-15
fix(probe-updates): spend the probe budget on the stalest entries first

Yur0918:
## What

`pending` was left in README order, so the nightly worked through the same list from the same end every time. This sorts it by `checkedAt` ascending instead:

```js
const pending = urls
  .filter((url) => !fresh(map[url]))
  .sort((a, b) => String(map[a]?.checkedAt ?? '').localeCompare(String(map[b]?.checkedAt ?? '')))
```

Entries with no `checkedAt` — never probed, or listed since the last run — have no date to sort on and land at the front, which is where they belong.

## Why

The nightly is the only run that probes with `PROBE_ALL`, and it cannot finish the list. It issues two API calls per repo (`/releases/latest` + `/commits?per_page=5`) across ~3,600 listed repos — about 7,000 calls — against the 1,000-request hourly budget this workflow's own comments already track (#1673, #3512). Once the budget is spent `gh()` throws, `probe()` catches it and returns `null`, and by design the entry keeps its previous value ("a failed repo keeps its old entry").

That fallback is right. What is wrong is the order: because it never changes, the same tail sits past the cut-off every single night and is never refreshed at all.

## What that looks like downstream

The `dsh-plugin-updates` payload published today (`2026.915.3405`) reports `updated: 2026-09-15`, yet 1,177 of its 3,477 entries were last checked on **2026-09-05**:

- 2026-09-14 — 1,266 entries
- 2026-09-05 — 1,177 entries
- 2026-09-15 — 1,033 entries

This is the root cause of dsh-market/dsh-market#614, which was reported as this script mangling UTF-8. It is not: `probe-updates.mjs` reads with `fetch` + `res.json()` and writes a Buffer, and `publish-updates.mjs` does the same — there is no single-byte codec anywhere on the path.

The reported entry, `mafeis/dsh-net-proxy`, was captured on 09-05 while its release body still held `?`, and its author corrected the body on GitHub on 09-11 (`updated_at`) — the frozen copy simply never went back for it. Across the whole payload only 12 entries show that collapse, and 11 of them still carry `?` upstream today, so those are author-side rather than ours.

## Measured

Replaying both orderings over the live README (3,632 listed repos) and the published payload (3,477 entries, 1,177 of them stale). With a 1,000-entry budget:

- README order (today): **1** of the 1,177 stale entries is reached
- `checkedAt` ascending: **819** of the 1,177 stale entries are reached

## Scope

One comparator and its comment. It does not change what is fetched, `PROBE_ALL`, the serial retry pass, the caches, or what counts as fresh — a run cut short just rotates the list across nights instead of starving one end of it. It needs no cache-version bump: `data/updates.json` keeps its shape.

## Verification

`node --check scripts/probe-updates.mjs` passes. I could not run the probe end-to-end — it needs `GITHUB_TOKEN` and a full nightly window — so the numbers above come from replaying the two orderings over the published data, not from a probe run.


Yur0918:
Superseded by #5149, which I had not seen when I opened this: same fix, six
hours earlier, and better — a tested helper (`lib/probe-order.mjs` +
`probe-order.test.mjs`, wired into the check step) instead of an inline
comparator. Two PRs for one comparator only spends maintainer attention, so I am
closing this one rather than asking anyone to choose.

The one thing this branch has that #5149 does not is a measurement, and I have
posted it there: replaying both orderings over the published payload (3,477
entries, 1,177 still on `checkedAt: 2026-09-05`) against a 1,000-entry budget,
README order reaches 1 of those 1,177 while `checkedAt` ascending reaches 819.
Counterfactual replay over the published data rather than a probe run — the run
log in #5149 is the direct evidence, this is corroboration.

Branch left in place in case #5149 stalls.

PR #5193, opened 2026-09-15
add drscrewdriver/dsh-arrowkey-nav (ui)

drscrewdriver:
- [x] I added **one file** at `data/plugins/<owner>__<repo>.yml` — that single file is the whole submission. The READMEs are regenerated on `main` after merge: don't edit them by hand, and you don't need to commit them either / 我新增了一个 `data/plugins/<owner>__<repo>.yml` 文件——**这一个文件就是全部投稿**。README 会在合并后由 `main` 自动重新生成：不要手工编辑，也不需要提交
- [x] My repo's `package.json` declares **`dsh.bundle`** (not just `dsh.client`) / 仓库 `package.json` 已声明 `dsh.bundle`（只有 `dsh.client` 无法安装）
- [x] My repo is at least **1 day old** / 仓库创建满 1 天
- [x] `category` is one of `agi ui usage theme model identity session memory tools wsl browser vision voice docs skill workflow git notify dev security remote market fun`, and themes/skins go under `theme` / `category` 取值正确，主题/皮肤类请用 `theme`
- [x] Description states what the plugin does, no superlatives / 描述只说功能，不带营销词
- [x] My repo has the `dsh-plugin` topic / 仓库已打 `dsh-plugin` topic

Add [drscrewdriver/dsh-arrowkey-nav](https://github.com/drscrewdriver/dsh-arrowkey-nav) to the **Interface & UX** category.

- url: https://github.com/drscrewdriver/dsh-arrowkey-nav
- category: `ui`
- description.en + description.zh included
- Entry file only — the READMEs are left for `sync-readme.yml` to regenerate on main
- Repo declares `dsh.bundle` (plus `dsh.client`), published on npm as `dsh-arrowkey-nav@0.1.1`
- `@deepseek-ai/*` packages are injected at runtime, not bundled; the package has **no runtime dependencies**

## What it does

Switch sessions without leaving the keyboard.

| Key | Action | Scope |
| --- | --- | --- |
| `↑` | previous session | inside the current workspace |
| `↓` | next session | inside the current workspace |
| `←` | previous workspace | all workspaces |
| `→` | next workspace | all workspaces |

`↑`/`↓` wrap at the ends of the workspace and never cross into a neighbouring one — that is what
`←`/`→` are for. After a switch the target row is scrolled into view.

**Identity comes from the controllers, never from the DOM.** Rows carry no `data-*` and no `id`,
so a row cannot be asked what session it is; switches go through `ctx.sessions.open(id)` and the
workspace order comes from `ctx.workspaces.list.getSnapshot()`.

**"Next" follows what the sidebar is actually drawing.** The sidebar reconciles its own persisted
order with membership and promotes recently-active sessions to the top, so an old session jumps to
the first row the moment you open it. The plugin reads the rendered row order and accepts a
section only when its rows account for every session the controller reports; an ambiguous or
mismatched section falls back to the controller order (which is also what happens when the sidebar
is collapsed to a rail) instead of guessing.

**It does not steal keys that belong to someone else.** Any modifier, IME composition, an event
already handled upstream, or a focus inside a dialog, menu, or another editable field leaves the
event alone. The listener registers in the capture phase but only calls `preventDefault()` when it
actually navigates.

**The composer is handled explicitly.** An empty composer yields its arrow keys — there is no caret
to move — which is what lets you navigate right after sending a message while the composer still
holds focus. Once the composer holds a draft, its arrow keys belong to the caret again, so editing
a prompt is never hijacked.

**Written in TypeScript**, no React components, no CSS, and no platform-module requests; the
browser half is generated into the module-loader contract the page expects. The test suite covers
the resolver, the key guards, display-order reconciliation and the built bundle. The README
documents its known limitations (collapsed groups and a rail-collapsed sidebar do not scroll)
rather than hiding them.

**DSH `>=0.1.2-rc.1 <0.2.0-0`** with Node.js 22.6+.

## Documentation

README, CHANGELOG and the installation guide each ship in English, Simplified Chinese, Japanese
and Korean.


drscrewdriver:
@fkysly  受累走个面 过下审

PR #5396, opened 2026-09-18
Add Pummelchen/TinyTitan#dsh-tinytitan

Pummelchen:
Adds `data/plugins/Pummelchen__TinyTitan--plugins-dsh-tinytitan.yml` — one file, per contributing.md. (An earlier attempt, #5094, was closed without a comment; this is the same entry, re-verified against the current `main`.)

What the plugin does: keeps a locally served TinyTitan model reachable from the harness. `src/route.js` with `src/generate.js` refreshes the `llm-pi-ai` route from the models the running server reports (`TinyTitanServer --catalog`), so no checkout is needed, and `src/compaction.js` mounts a compaction backend that runs with reasoning off, so a summariser's output cap is not spent on thinking.

Entry requirements:
- `plugins/dsh-tinytitan/package.json` declares `dsh.bundle.patch` → `./cordis.patch.yml`, which sits beside it.
- Real, working code: plain ESM, MIT, 66 `node --test` cases (65 pass, 1 skipped — a shell byte-equality check that needs a built server).
- The repository is older than a day and carries the `dsh-plugin` topic.
- `category: model` — the plugin configures the model route and the compaction call; it ships no UI.

Checked with the catalogue's own tooling against a fork level with `main`: `validateEntries()`, `node scripts/generate-readme.mjs` (both READMEs regenerate), `npx awesome-lint`, and `node scripts/build-site.mjs` with the READMEs regenerated — entry and detail page present in both locales.


Pummelchen:
Both checks are green and the branch is level with `main`, so this is waiting on a merge rather than on a fix. Could a maintainer take a look when convenient? If anything about the entry or the plugin needs changing first, say so and I will adjust it.

PR #5420, opened 2026-09-19
Add iimaguest/dsh-sandbox-escalation-guard

iimaguest:
## What this adds

`iimaguest/dsh-sandbox-escalation-guard` — category `security`.

It stops the sandbox-escalation fields being offered to a model when the calling
session can never grant them, which is what makes **GPT-family models fail every
tool call** in a `danger-full-access` session.

## The defect

`bash`, `write` and `edit` advertise `sandbox_permissions` with
`enum: ["workspace-write", "danger-full-access"]`, but `approveEscalation`
accepts a level only when it is *strictly wider* than the one already in effect.
A session at `danger-full-access` is therefore handed two values that are **both
rejected** — one a reduction, the other the mode already in force — while the
prompt urges the model to use the field. The two errors that result never name
the actual repair, which is to omit both fields.

## Why it reads as a model-specific bug

GPT-family models complete every optional property in a tool schema. The two
optional properties on those tools are exactly the two a full-access session can
never use, so GPT fills them in on every call and every call dies before
anything runs. A model that sends only what a call needs never touches them.

From one live session, same schema, same repo, same mode:

| model | tool calls | emitted `sandbox_permissions` | bash calls | bash + optional field |
|---|---|---|---|---|
| `zai/glm-5.3` | 211 | **0** | 86 | 48 |
| `oauth-codex/gpt-6-astra` | 10 | **8** | 8 | **8** |

The GLM column is not a model that tried and succeeded — it emitted the field
zero times in 211 calls. The wire-level `request/header` for that session
confirms the field really was published to the model, with that enum.

## What it does

Narrows the published enum **per request** at `system-prompt/assemble` against
the calling session's effective mode, and removes both the field *and* its
escalation prose when no level is grantable. A cached or non-conforming call is
denied at `tools/pre-execute` with text naming the repair. It never widens
anything: a mode the harness would have refused is simply no longer offered.

The ladder is the one `approveEscalation` enforces, mirrored with a drift-guard
test that fails if the harness changes it. **Zero runtime dependencies.**

28 tests, including what the model is actually handed per mode and that
unmounting restores the original schema exactly. `npm run verify` reproduces the
per-mode output.

## Overlap with `apex-mochen/dsh-sandbox-arg-guard`

These address the same defect by different means, so I want to be upfront rather
than let a maintainer find it:

- **`dsh-sandbox-arg-guard`** listens on `tools/execute`, and *after* the
  documented rejection re-issues the identical call once without the argument.
  Recovery. Its own description notes that DSH's source calls the check
  "deliberately not a schema constraint".
- **This plugin** stops the un-grantable value being published, so the model
  never forms the bad call. Prevention. The prompt instruction it removes
  ("escalate immediately … retry with `sandbox_permissions`") is the same text
  that makes a schema-completing model emit the field, so removing the field
  without the prose would leave the model being told to use something that is no
  longer there.

Worth noting `dsh-sandbox-arg-guard`'s mechanism works on *any* redundant value,
not only ones the session cannot grant, so it is not made redundant by this. If
you consider one sufficient for the list, that is your call — I am flagging it
rather than arguing for both.

## CI: the 1-day age gate

The repository was created at `03:17Z` and this PR follows shortly after, so the
age check will fail on first run. Per `contributing.md` that clears by itself —
no resubmission, push or reopen from me, and `regate.yml` will flip it. I am
opening now so the review can start rather than waiting on a timer.

## Repo

- `package.json` at the root declares `dsh.bundle` → `cordis.patch.yml`
- topic `dsh-plugin` set
- `screenshots.json` declares four images, generated from verified output
- Apache-2.0

Install: `dsh plugin --profile web add github:iimaguest/dsh-sandbox-escalation-guard` — verified working from GitHub.


iimaguest:
Closing this. I opened it before the 24-hour window I had committed to, which was my error — the repository age gate on this submission is not something to open early and wait out. I will reopen it after the window has genuinely passed. No review time needed on this one.

PR #5483, opened 2026-09-20
add(fu827707013): dsh-model-health-probe (manual model probe in the s…

fu827707013:
Adds `fu827707013/dsh-model-health-probe` to `dev`.

**What it does:** a session-view tab where you pick provider → API type → model → route, then send one real raw HTTP request over `openai-completions` / `openai-responses` / `anthropic-messages` (streaming optional). One screen shows latency, TTFT, HTTP status, token usage, the response body, and a three-layer diagnosis chain (API request / model response / strict check). Errors are rendered in plain language.

**Why it exists:** the same provider often carries several protocols at once, and the `/v1` suffix rule is *opposite* between them — `openai-completions` and `openai-responses` need `/v1` in the baseURL, `anthropic-messages` must not have it (the SDK appends it). A misconfigured baseURL surfaces as a bare 404 or 401, which says nothing about which layer broke. The plugin sends a minimal protocol request over raw HTTP (not via the host's `ctx.llm.stream`, so no `compat` / `headers` / `retryPolicy` are applied) and reports per-layer conclusions.

**Scope:** it is a manual probe, not a monitor — no polling, no alerting, no history, no state files. Records live in an in-memory ring buffer only. The only write is the "set as default" button, which writes the `agent-default-model` section and nothing else.

**Gate checks:**
- `dsh.bundle` declared: `{"dsh": {"bundle": {"patch": "./cordis.patch.yml"}}}` ✅
- repository age: created 2026-09-18 ✅
- not archived, not the harness itself, not a first-party package ✅
- one entry in this PR ✅

**Verification:** `npm test` 259/259 and `npm run acceptance` 110/110 pass; the acceptance run also scans for credential leakage across four surfaces (response, logs, state files, errors) and reports zero hits. The published tarball's `lib/` matches the repository byte for byte.


fu827707013:
两项检查都已通过（`PR check` 7 分钟、`Submission gate` 报 "Entries look good"），无冲突，可以干净合并。谢谢！

Both checks are green (`PR check` in 7m, `Submission gate`: "Entries look good") and there are no conflicts — ready to merge. Thanks!

PR #5485, opened 2026-09-20
Add dsh-todo-dock (ui)

vanddccd:
Adds [dsh-todo-dock](https://github.com/vanddccd/dsh-todo-dock) — a rebuilt todo dock for the web UI.

**What it does**
- Takes over the built-in todo dock by substituting the `conversation.input.dock` entry with `id: 'todo'` (no dsh source changes; the original component is restored on dispose).
- Tri-segment progress rail: completed (green) / in progress (blue) / pending (grey). The rail is hidden entirely when nothing has started, so a bare grey line never reads as a broken progress bar.
- Finished items collapse into a single summary row (click to expand); when everything is done the panel collapses to one row.
- Generic by construction — the component reads only `{ content, status }`, never parses the task text, and groups by `status` alone. Checked by rendering the same component against 10 datasets (development steps, video review, data backtesting, translation, a long CI session, English-only, a single item, over-length text, 22 items, and empty).
- Reuses the host's existing locale keys, so no new dictionary ships. Colours come from host `--dsw-*` tokens, so light and dark both work.

**Motion** (transitions.dev motion-token scale)
checkbox check (the tick draws using `path.getTotalLength()`, never a hard-coded dasharray) · icon swap (three status icons stacked in one cell) · accordion (`grid-template-rows 0fr ↔ 1fr`, chevron flips via `scaleY`) · number pop-in (the counter re-enters per digit, with the required forced reflow) · card resize (progress segment widths) · texts reveal (row entry). Each snippet keeps its own `prefers-reduced-motion` guard.

**Requirements**
`dsh.bundle` manifest + `cordis.patch.yml` at the root, real working code (the client bundle is built and committed), and the `dsh-plugin` topic is set.

**Verification** (reproducible)
- `test/harness.html` loads the built `client.js` under the dsh module-table contract: 7/7 PASS — module id, `apply` export, slot entry replaced, stylesheet injected, motion tokens injected, effect registrations, takeover held.
- `test/index.html` renders the real component source with React 18 across the 10 states listed above.

Note: the repo was created today, so the 1-day age check will fail on this run. Per `check-submission.mjs` the check re-runs itself, so I'm leaving the PR open instead of resubmitting.


vanddccd:
Withdrawing this submission for now. I'd rather run the plugin in daily use for a while before listing it publicly, and the repository age check would need to clear anyway. The submission tooling was clear enough that the failing check was unambiguous — thanks for that. I may reopen later if it holds up in practice.

(推送历史已重写：撤下了一份包含私有项目工作流细节的设计稿归档。)

PR #5529, opened 2026-09-20
Add dsh-audio-cue

tangjx2007:
Adds dsh-audio-cue: audio cues for approval and question prompts, turn completion, background jobs, failures and sandbox escalations.


tangjx2007:
Adds dsh-audio-cue: audio cues for approval and question prompts, turn completion, background jobs, failures and sandbox escalations.

PR #5532, opened 2026-09-20
Add HorusJiang/dsh-jev-tools

HorusJiang:
Adds one entry file: `data/plugins/HorusJiang__dsh-jev-tools.yml`.

- **Repository:** https://github.com/HorusJiang/dsh-jev-tools (public, MIT, published on npm as `dsh-jev-tools`, currently 0.1.2)
- **Installable:** `package.json` declares `dsh.bundle.patch` → `./cordis.patch.yml`, the repo ships that patch, and the entry id is `jev-tools`; the `dsh-plugin` topic is set. The bundle mounts one host plugin plus its Web settings page.
- **Category `tools`:** it acts on tool results — it prunes oversized `read` / `grep` / `glob` / `web_fetch` / `web_search` output at `tools/post-execute` — and it registers two tools of its own, `jev_ask` and `jev_gate`.

**What it does.** [Jev](https://typesafe.ai) (TypeSafe's System One model) answers typed questions instead of generating text. This plugin runs three judgments automatically during a session:

1. prune oversized tool output down to the segments relevant to the request, keeping deterministic head/tail floors and failing open on every error path;
2. screen fetched pages for instructions aimed at the model — advisory only, it never blocks a call or rewrites content;
3. pick which skill fits the next step, at most one suggestion per turn;

plus two tools the model can call: `jev_ask` for arbitrary typed questions, and `jev_gate`, a pre-completion check that escalates every unclear answer instead of approving it. Each judgment is appended to a ledger that survives restarts, and `/jev-status` reports what has been judged so far.

**Key.** It reads `TYPESAFE_API_KEY` (the official SDK's variable). With no key it mounts and does nothing — no request of any kind is sent, and `/jev-status` says so. The key can also be pasted into the plugin's settings page, which stores it through the host's credential store.

**Note on the gate.** The `check` job passed; the submission gate fails on one thing only, the age bar — the repository was created a few hours before this PR. Per the gate's own message the verdict re-runs by itself and clears in about 24h, so I am leaving the PR alone rather than reopening it.


HorusJiang:
Status note, since the only red check here is the repository age gate: everything else on this entry passed (`check`, 7m40s).

The repository was created at `2026-09-20T16:32:55Z`, so the 1-day bar is satisfied from `2026-09-21T16:32:55Z` (local `2026-09-22 00:33`, UTC+8). Per the gate's own message, the verdict re-runs by itself and I am leaving the pull request alone — no push, no reopen.

The entry file itself is unchanged since it was opened. The package has moved on to `0.1.5` (same `dsh.bundle.patch` → `cordis.patch.yml`, entry id `jev-tools`); the changes since the submission were documentation, plus a `dshWorkshop` metadata block that only third-party catalogs read.

PR #5545, opened 2026-09-21
Add pearjelly/deep-blend — Blender 3D animation workbench for DSH

pearjelly:
Adds one entry, in `tools`.

**What it does.** A Blender 3D animation workbench: SceneSpec is the source of truth and `.blend` is a compiled artifact. Sixteen model-visible tools cover scene specs, previews, visual review and delivery renders. Every accepted change is one immutable revision published with a single `rename`; anything above a configured frame count does not render a frame without an explicit approval.

**Install** (documented in the README, and the form this entry's `url` implies):

```sh
dsh plugin --profile web add 'github:pearjelly/deep-blend#path:/packages/deepblend/bundle'
```

The four sibling packages resolve through the same repository (`github:pearjelly/deep-blend#path:/packages/deepblend/…`), so the source route needs no registry packages. Verified end to end in a throwaway `DSH_HOME`: pnpm installs seven packages, the profile composes all four rows, and the preset roster reads both presets as healthy.

**On the two things the review checklist asks about most:**

- **Not a meta-package.** The patch inserts four rows and every one of them carries a `config:` block; the fourth mounts a deployer that ships the agent presets and installs them into `<DSH_HOME>/.agent-presets/`. That row is what makes one install cover both planes — the sixteen tools belong to an agent preset, not to the host composition, so a bundle that only mounted the host would install and leave every session unable to render. `dsh-expert-orchestrator` is the same shape. The repository's `contract/contributor-surface.test.mjs` fails if a row ever composes without configuration.
- **Nothing alarming at install time.** Importing every package writes nothing and opens no handle — asserted by running each import in a child process with throwaway `cwd`/`HOME` and comparing the directories. The one write this plugin makes is at composition, into its own preset ids under the preset root, and a home it cannot write is logged rather than thrown.

**Claims checked against the code before submitting.** The repository's `contract/listing-entry.test.mjs` reads this same entry and checks it: the url's subdirectory exists and declares `dsh.bundle`, the category is one of the ecosystem's ids, `description.en` is a single line whose tool count is compared with the roster the tools are declared in (16), and each behavioural claim (immutable revisions, an approval gate) is tied to a numbered decision in the repository's register. `screenshots.json` sits beside the bundle's `package.json`.

The `dsh-plugin` topic is on the repository.

---

新增一条，归入 `tools`。

**它做什么。** 一个 Blender 3D 动画工作台：SceneSpec 是事实来源，`.blend` 是编译产物。16 个模型可见工具覆盖场景规格、预览、视觉评审与交付渲染。每次被接受的修改都是一个不可变 revision，用一次 `rename` 原子发布；超过配置帧数的渲染在没有明确审批之前一帧都不渲。

**安装**（README 里写的，也就是本条 `url` 隐含的形式）：

```sh
dsh plugin --profile web add 'github:pearjelly/deep-blend#path:/packages/deepblend/bundle'
```

四个兄弟包通过同一个仓库解析（`github:pearjelly/deep-blend#path:/packages/deepblend/…`），所以源码路线不需要任何 registry 包。已在一个临时 `DSH_HOME` 上端到端实测：pnpm 装上七个包、profile 组合出全部四行、preset roster 读出两个 preset 且健康。

**关于评审清单最关心的两点：**

- **不是纯聚合包。** patch 插入四行，每一行都带 `config:` 块；第四行挂载一个 deployer，它随包携带 agent preset 并把它装进 `<DSH_HOME>/.agent-presets/`。这一行正是「一次安装覆盖两个平面」的原因——16 个工具属于 agent preset 而不属于宿主组合，只挂宿主的 bundle 会装上却让每个会话都渲不出东西。`dsh-expert-orchestrator` 是同一形态。仓库里的 `contract/contributor-surface.test.mjs` 会在任何一行缺少配置时报红。
- **安装期没有可疑行为。** 逐个 import 每个包都不写文件、不打开句柄——这一条由一个子进程实测（临时 `cwd`/`HOME`，前后比对目录）。这个插件唯一的写入发生在组合时，写进 preset 根下它自己的 preset id；写不进去的 home 只记日志，不抛异常。

**提交前已对着代码核过声明。** 仓库里的 `contract/listing-entry.test.mjs` 读的就是这一条并逐项检查：url 的子目录存在且声明了 `dsh.bundle`、分类在生态的 id 集合内、`description.en` 是单行且其中的工具数与工具声明处的花名册一致（16）、每条行为声明（不可变 revision、审批闸门）都对应仓库决策登记里一个编号决策。`screenshots.json` 就在 bundle 的 `package.json` 旁边。

仓库已加 `dsh-plugin` topic。


pearjelly:
## A third install route: a self-contained release tarball

This branch now declares a `tarball` alongside the source route. Both are measured on a scratch `$DSH_HOME`, not asserted from inspection.

**Source** — `dsh plugin --profile web add 'github:pearjelly/deep-blend#path:/packages/deepblend/bundle'`

```
Packages: +7        # the bundle and its six siblings, all through this repository
Done in 11.8s
```

**Tarball** — `dsh plugin --profile web add https://github.com/pearjelly/deep-blend/releases/latest/download/deepblend-bundle.tgz`

```
Packages: +1        # it fetches the artifact and resolves nothing
Done in 2.3s
```

Both then behave identically: the four composed rows, `/deepblend/capabilities` HTTP 200 with `hostApiVersion 4`, and both agent presets deployed into `<DSH_HOME>/.agent-presets/` byte-identical to the repository's `deepblend/presets/`. DSH's own `discoverPresets` reports `problem: null` for both.

### Why the tarball is self-contained, and how

Not a bundler. The artifact carries the six `@deepblend/*` packages inside its own `node_modules` and lists them under npm's `bundledDependencies`; pnpm uses the copies in the tarball and never resolves the specs. Measured, because the design rests on it: it still does this when a bundled dependency's spec is a version that exists nowhere.

### The asset name has no version in it, on purpose

`/releases/latest/download/` takes the filename literally, so `…-0.1.0.tgz` is a URL that works on the day it is written and 404s at the next release — the pattern `scripts/probe-tarballs.mjs` warns about. Every release attaches the same version-free name; the version lives in the tag.

### Where the entry comes from

The file in this PR is generated. `deepblend/docs/listing-entry.yml` is the source, and the submitted file is its projection — the source from its first permitted key to the end — so the description, the URL and the tarball URL are edited in exactly one place and cannot drift. The repository asserts that, including a case that fails if the description is ever written into a second file.

Verified against this repository's own validator before pushing: `validateEntries` over the full 4063-entry set reports no problems, and `tarballProblem` accepts the URL.

pearjelly:
## Against the review checklist

`contributing.md` names eight things a maintainer looks at. Most are answered by the code; two of them a reader cannot get from the entry, so here they are with the measurements behind them.

### (4) Is it already covered

There is a Blender entry on the list: `CheshireJCat/blender`, category `tools` — 30 modeling/reconstruction skills, 13 runtime tools, 26 deterministic helpers. **Same domain, not the same plugin**, and the guide's rule is whichever is better rather than whoever got here first. What this one does that the other does not, each asserted in the repository:

- **SceneSpec is the source of truth; `.blend` is a compiled artifact.** The scene is a declarative spec that compiles deterministically — same spec, same digest, checked by re-compiling stored revisions.
- **Immutable revisions.** Every accepted change is one revision: validated, compiled in a staging directory, verified, then published by a single `rename`. A failure never touches the current revision.
- **A score the model cannot write.** The visual loop renders a contact sheet, and a deterministic scorer measures occlusion and framing *from the pixels*. The model's own findings are recorded separately, and a fix is adopted only if the score really improved.
- **An approval gate on cost.** Above a configured frame count, not one frame renders without an explicit approval naming the revision it will write.
- **Renders that survive being killed.** Frames on disk are the truth, not a counter — after a `SIGKILL` the next process stops the orphaned renderer, rebuilds the ledger from the frames themselves, and finishes. Measured: a delivery killed at frame 6 of 60 completed all 60 and encoded a 1920×1080 `final.mp4`.

### (7) Is it a meta-package

This is the item that could send the submission back, so it is answered by the market's own second CI check rather than by an argument:

> **`dsh.bundle`** — fetched from your repo's `package.json` (root, or a `packages/` · `plugins/` · `apps/` subpackage). Declaring only `dsh.client` fails here.

**Measured: exactly one package in this repository declares `dsh.bundle`, and it is the one this entry points at.** So "list the plugins, not the bundle" has no alternative here — there is no other package the gate would accept. The six siblings are internal modules of *one* product (contracts, the Blender execution seam, the business facade, the browser host half, the tool plane, the preset deployer), not other entries. This entry points at no other entry's work, so the double-counting the rule exists to prevent cannot occur.

The honest part, since you may go and look: the patch's `config:` blocks set 18 keys across four rows and **none of them differs from the package's own schema default**. They make the composed configuration explicit in `--dump-config` rather than change it. What the patch contributes is *which* rows mount — and without it none of them does, because a dependency list alone mounts nothing. The bundle is the composition, which is what a DSH Host Bundle is (SPEC §4.1, §18).

### (7) continued — this shape is already on the list, eight times

The argument above is about what the gate would accept. Here is what the list already contains, read from `data/plugins/*.yml`: **eight entries point at a `packages/bundle` directory**, across six categories (`ui`, `security`, `model`, `market`, `theme`, `session`). Two of them, opened directly:

| repository | what `packages/bundle` holds | patch rows | dependencies |
|---|---|---|---|
| `ayahunter/dsh-trail` | `cordis.patch.yml` (**421 B**) + `package.json` — no source at all | **1** | **1** |
| `f-infinite-z/dsh-plugin-ops` | `cordis.patch.yml` (285 B) + manifest + `src/`, `tests/` | **1** | **1** |
| **this submission** | `cordis.patch.yml` (11.9 KB) + `lib/` + `screenshots.json` | **4** | **6** |

Both accepted bundles are composition wrappers — one row, one dependency — and neither is a meta-package of *other people's* plugins. This submission is strictly more than either by every one of those measures, and its six dependencies are its own repository's internal modules rather than other entries.

### (6) Does the PR touch anything else

No. One file added, nothing modified — checked through the API rather than the diff.

### Also worth flagging before you find it

The bundle mounts a deployer row that **writes to `<DSH_HOME>/.agent-presets/` when the profile composes**. That is install-time behaviour, so it should be stated rather than discovered:

- it writes only `<DSH_HOME>/.agent-presets/<preset-id>/` — never the root itself, never another preset;
- it is idempotent, and removes only files an earlier release of *this* package left behind;
- a home it cannot write is **logged, not thrown** — a convenience copy must not take a deployment down;
- `deploy: false` leaves an operator-managed preset root untouched.

It exists because the 16 model-visible tools belong to an agent preset, and `dsh plugin add` does not write a user's preset root — without it the plugin would install, serve the workbench, and leave every session unable to render anything.

### (5) Anything alarming in the source

Measured rather than asserted:

- **No credential access anywhere in the shipped packages.** The only environment variables read are `PATH`, `HOME`, `TMPDIR` and `DSH_HOME` — locating executables and the DSH home. No API keys, no tokens, no `.npmrc`.
- The one module that mentions credentials *removes* them: `contracts/lib/redact.js` strips userinfo, query string and fragment from a URL before it is logged or returned, and says what it removed.
- **Two outbound fetch sites.** The asset ingest in the host (bounded redirect hops, a timeout, and an approval boundary for network sources), and the workbench client, which fetches only its own relative routes. Nothing else in the product opens a socket.
- No build step and no install scripts: `npm pack` on each package produces `lib/` (plus `python/` for the Blender seam, `presets/` for the preset package) and the manifest.

pearjelly:
## All three install routes now work, and the npm one is the fast path

Since the last comment the seven packages were published to npm, so the entry no longer only claims a source route. Every route below was measured on a scratch `$DSH_HOME` with the same criteria — four composed rows, `/deepblend/capabilities` HTTP 200 with `hostApiVersion 4`, both agent presets deployed byte-identical to the repository's `deepblend/presets/`, and DSH's own `discoverPresets` reporting `problem: null` for both.

| route | command | what pnpm fetches |
|---|---|---|
| **npm** | `dsh plugin --profile web add @deepblend/dsh-blender-bundle` | **7, all from the registry** |
| source | `dsh plugin --profile web add 'github:pearjelly/deep-blend#path:/packages/deepblend/bundle'` | 7, all through this repository |
| tarball | `dsh plugin --profile web add …/releases/latest/download/deepblend-bundle.tgz` | **1** — the artifact carries its siblings inside itself |

The npm route is worth one sentence more, because "resolved from the registry" is not automatic. This repository's manifests carry self-referential `github:` dependency specs, which the **source** route needs — a remote install fetches the bundle and then resolves its dependencies, and a version pin resolves to nothing when the siblings are unpublished. Publishing those manifests as-is would have made the npm route a registry install in name only: pnpm would take the bundle from the registry and go straight back to git for the six siblings. The publish tool rewrites the specs to exact versions in a staged copy, and the published manifests were read back to confirm it — `npm view @deepblend/dsh-blender-bundle dependencies` answers six `0.1.0` pins and no git spec.

The list harvests the npm mapping itself rather than reading it from the entry, so the harvest condition was checked with your own logic: the name is read from the manifest at this entry's URL, the registry has it, and its `repository` field contains `pearjelly/deep-blend` — so `probe-npm.mjs` would record `{ npm: '@deepblend/dsh-blender-bundle', version: '0.1.0' }` for this entry on its next pass.

Nothing in the entry file changed for this; only its own analysis header, which is not part of the submission.

PR #5143, opened 2026-09-15
Add hoyyang/dsh-concise to the Session category

hoyyang:
Add [hoyyang/dsh-concise](https://github.com/hoyyang/dsh-concise) to the **session** category.

- url: https://github.com/hoyyang/dsh-concise
- category: `session`
- description.en + description.zh included
- READMEs regenerated (3707 entries)
- Published on npm as `dsh-concise@0.7.0` (public), installable via `dsh plugin add hoyyang/dsh-concise`; GitHub Release prebuilt tgz also declared via the optional tarball field

PR #5550, opened 2026-09-21
Re-add hoyyang/dsh-concise to the Session category (v0.10.6)

hoyyang:
Re-adds the dsh-concise listing. The previous PR #5143 was merged on 2026-09-18 but the entry was lost when the catalog was regenerated (the merge commit is not an ancestor of current main; the yml is absent from the tree).

- tarball: dsh-concise v0.10.6 (GitHub Release asset, sha256 b1dcc5fed0b5)
- latest: digest card deliverables zone (artifact files as clickable chips with OS open), compliance self-check loop, CJK path support
- READMEs regenerated from data/plugins via scripts/generate-readme.mjs


hoyyang:
Superseded: the entry was re-landed upstream by the conflict-resolution commit; this PR is redundant. Follow-up bump: the v0.10.6 tarball PR.

PR #5634, opened 2026-09-21
Add slow-stack/euthyna

modusensus:
Adds [slow-stack/euthyna](https://github.com/slow-stack/euthyna) under Skills.

euthyna (εὔθυνα — the audit Athenian officials submitted on leaving office) is
a security-audit discipline for AI coding agents, shipped as a dsh bundle:
`plugin/index.js` mounts the packaged skill on `ctx.skills` through the official
`FileSystemSkillProvider`, and `package.json` declares `dsh.bundle.patch`
pointing at `cordis.patch.yml`, which inserts the `euthyna` row.

The companion CLI (npm: [euthyna](https://www.npmjs.com/package/euthyna), zero
dependencies) produces the deterministic facts the skill adjudicates:

- `euthyna history` — attributes every deleted line to a commit, classifies it
  from its message, its own diff and the deleted line's content, flags
  deletions that came from security fixes, and with `--origins` chases the
  commit that first introduced the content.
- `euthyna coverage` — whether a changed symbol was ever invoked by a test; it
  never reports "executed" (V8 coverage marks unreachable code as covered),
  only never-invoked or unknown.
- `euthyna gate` — mechanically checks each adjudication finding against six
  gates and downgrades whatever cannot produce evidence to an observation.

The skill is user-invocable only (`disable-model-invocation: true`): an audit
runs when the user asks for one. Every claim above is checkable in the
repository (README, `src/`, `plugin/`).


vecnode:
https://github.com/vecnode/vn-harness

PR #5636, opened 2026-09-21
Add BOWLUNA/dsh-zcode-git

BOWLUNA:
Adds one entry: `data/plugins/BOWLUNA__dsh-zcode-git.yml`.

**What it is.** Six git tools for the agent — `git_status`, `git_diff`,
`git_log`, `git_branch`, `git_commit`, `git_stash` — so it stops driving git
through `bash`. The package declares `dsh.bundle.patch`.

**Why the entry is worded the way it is.** The description makes three claims,
and each is checkable against the code:

- *ten `git -c` overrides* — the pinned list is `PINNED_CONFIG` in `src/exec.js`
  (colour, pager, quotepath, relative paths, log date/signature, diff prefixes and
  external diff, detached-HEAD advice). They exist because each one changes the
  bytes the parsers read.
- *every invocation is an argv array, so no shell ever parses an argument* —
  `src/exec.js` calls `ctx.subprocess.spawn` with an argv array and never builds a
  command line. `test/e2e.test.js` stores a commit message containing `%PATH%`,
  `$(...)`, quotes and a newline, then reads it back out of the repository.
- *mutations go through the approval service, refusing rather than proceeding* —
  `git_commit` and the mutating actions of `git_branch`/`git_stash` call
  `ctx.approval.request()`; with no approval service mounted the mutation aborts.

**Deliberately absent**: `push`, `fetch`, `pull`, `reset`, `clean`, `rebase`,
`remote` and config writes — an operation that can destroy a working tree in one
call is not something this plugin wants to expose as a tool.

95 tests across six layers, including end-to-end runs against a real git binary
on Windows and Linux, and schema checks covering both `0.1.5-rc.2` and
`0.1.6-alpha.2`.

The pull request adds one file and touches no other entry.


vecnode:
https://github.com/vecnode/vn-harness

